PR #8, opened 2018-05-10
[CO 280] Switch to universum

HirotoShioi:
## Description

Migrate the code to use [universum](https://hackage.haskell.org/package/universum)

## Linked Issues

https://iohk.myjetbrains.com/youtrack/issue/CO-280

Branched from [this PR](https://github.com/input-output-hk/log-classifier/pull/7)

## Note

I did follow the migration guide written on the package document but I might have missed something. Please check carefully!

### DIFF
--- a/src/LogAnalysis/KnowledgeCSVParser.hs
+++ b/src/LogAnalysis/KnowledgeCSVParser.hs
@@ -4,18 +4,17 @@ module LogAnalysis.KnowledgeCSVParser
        ( parseKnowLedgeBase
        ) where
 
-import           Control.Applicative
-import           Data.Attoparsec.Text.Lazy
+import           Universum
+
+import           Data.Attoparsec.Text.Lazy as ALT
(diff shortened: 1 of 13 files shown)


src/LogAnalysis/KnowledgeCSVParser.hs
```diff
@@ -47,20 +46,20 @@ parseErrorCode =
...
 
--- |Parse each csv records
-parseKnowledge :: Parser Knowledge -- not really clean code..
+-- | Parse each csv records
+parseKnowledge :: Parser Knowledge  -- not really clean code..
 parseKnowledge = do
     e <- quotedField
-    _ <- char ','
-    _ <- char '"'
+    () <$ char ','
```
ksaric: Revert this, it's really not that readable. There are two options:
- use the previous version - `_ <- char ','`, which I prefer
- use `void` - `void $ char ','`

PR #13, opened 2018-05-23
[CO-296]  Extract and make ZenDesk API testable

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CO-296

A first major refactoring. I tried to pick the simplest way of introducing mocking capabilities, required for https://iohk.myjetbrains.com/youtrack/issue/CO-289


src/Lib.hs
```diff
@@ -1,8 +1,283 @@
...
 
 import           Universum
--- TODO(ks): Here we want to import the functionality we require in cardano-report-server.
-someFunc :: IO ()
-someFunc = putTextLn "someFunc"
+
+import           Data.Text (stripEnd, isInfixOf)
+import           Data.Attoparsec.Text.Lazy (eitherResult, parse)
+
+import           LogAnalysis.Types
```
HirotoShioi: Can you make these imports explicit?

src/Zendesk/Functions.hs
```diff
@@ -0,0 +1,174 @@
...
+              Just nextUrl -> go (list' <> pagen) nextUrl
+              Nothing      -> pure (list' <> pagen)
+
+    (TicketList page0 nextPage) <- liftIO $ apiCall parseTickets req
+    case nextPage of
+        Just nextUrl -> liftIO $ go page0 nextUrl
+        Nothing      -> pure page0
+
+-- | Send API request to post comment
+postTicketComment
```
HirotoShioi: Since we've defined `ZendeskResponse` in the previous PR, we can make this function to take `ZendeskResponse` as an argument which will be much more simpler
`postTicketComment
    :: (MonadIO m, MonadReader Config m)
    => ZendeskResponse
    -> m ()`

src/Zendesk/Types.hs
```diff
@@ -34,11 +119,16 @@ data Attachment = Attachment
...
+data RequestType
+    = Requested
+    | Assigned
+
 -- | The response for ZenDesk.
 data ZendeskResponse = ZendeskResponse
-    { zrComment     :: Text
-    , zrTags        :: [Text] -- TODO(ks): This should be wrapped
-    , zrIsPublic    :: Bool
+    { zrComment  :: Text
```
HirotoShioi: Can you add `!` to these fields?

src/Lib.hs
```diff
@@ -1,8 +1,283 @@
...
+        Left  err     -> error err
+    let cfg' = defaultConfig
+                   { cfgToken = stripEnd token
+                   , cfgAssignTo = assignTo
+                   , cfgKnowledgebase = knowledges
+                   }
+    let getAgentId = zlGetAgentId . cfgZendeskLayer $ cfg'
+    agentId <- runApp getAgentId cfg'
+    let cfg = cfg' { cfgAgentId = agentId }
+    -- At this point, the configuration is set up and there is no point in using a pure IO.
```
HirotoShioi: I'm not sure what this means (Not enough haskell knowledge to understand it).  Can you explain it to me?

src/Zendesk/Types.hs
```diff
@@ -1,28 +1,113 @@
...
+tokenPath = "./tmp-secrets/token"
+
+-- | Filepath to assign_to file
+assignToPath :: FilePath
+assignToPath = "./tmp-secrets/assign_to"
+
+-- | The Zendesk API interface that we want to expose.
+-- We don't want anything to leak out, so we expose only the most relevant information,
+-- anything relating to how it internaly works should NOT be exposed.
+data ZendeskLayer m = ZendeskLayer
```
HirotoShioi: I can see what it does but I'm not sure how this would enable us to do simple tests. (Again, now enough haskell knowledge!)

src/Lib.hs
```diff
@@ -1,8 +1,296 @@
...
+import           Data.Text (stripEnd, isInfixOf)
+import           Data.Attoparsec.Text.Lazy (eitherResult, parse)
+
+import           LogAnalysis.Types
+import           LogAnalysis.Classifier
+import           LogAnalysis.KnowledgeCSVParser
+import           Util
+import           Config
+import           CLI
+import           Zendesk
```
HirotoShioi: Can you make these imports explicit?

src/Lib.hs
```diff
@@ -1,8 +1,283 @@
...
+
+import           Data.Text (stripEnd, isInfixOf)
+import           Data.Attoparsec.Text.Lazy (eitherResult, parse)
+
+import           LogAnalysis.Types
+import           LogAnalysis.Classifier
+import           LogAnalysis.KnowledgeCSVParser
+import           Util
+import           CLI
+import           Zendesk
```
HirotoShioi: Can you make these import explicit?

ksaric: Yes, good catch.

src/Zendesk/Types.hs
ksaric: Yes, nice catch.

src/Zendesk/Functions.hs
ksaric: Nice, thanks.

src/Zendesk/Types.hs
ksaric: I guess you can imagine `ZendeskLayer` as an object (if an object is a dictionary of functions) that we pass around.
When we want to use it we use the `Reader` to fetch the function we are interested in.
The enables us to replace the functions anyway we want because they are defined in the `Config` that `Reader` is using.
In other words, if we now run a test and want `listTickets` to return an empty list, we can, because all we need to do is construct a new configuration with that function a simple constant that returns an empty list and we can test the rest of the code.
If this is confusing, the next PR might make more sense since it will deal with tests themselves.

src/Lib.hs
ksaric: It just means that there is no point in using `IO` all over the codebase since the `Config` is initialized and we require `Config` to use the `App` newtype.
The consequences of this are - try to see how many `runApp` were removed...

PR #14, opened 2018-05-30
[TSD-18] Add simple tests

ksaric:
Issue migrated from CO-289 to https://iohk.myjetbrains.com/youtrack/issue/TSD-18.

Simple tests working, still plenty of work to be done. Hopefully, this should persuade people that using tests leads to better code.

### DIFF
--- a/src/Lib.hs
+++ b/src/Lib.hs
@@ -57,7 +57,7 @@ runZendeskMain = do
     case args of
         CollectEmails            -> runApp collectEmails cfg
         (ProcessTicket ticketId) -> void $ runApp (processTicket ticketId) cfg
-        ProcessTickets           -> runApp processTickets cfg
+        ProcessTickets           -> void $ runApp processTickets cfg
         FetchTickets             -> runApp fetchTickets cfg
         ShowStatistics           -> runApp showStatistics cfg
 
@@ -101,12 +101,7 @@ processTickets :: App ()
 processTickets = do
     sortedTicketIds     <- listAndSortTickets
 
-    zendeskResponses    <- forM sortedTicketIds $ \ticketInfo -> do
-        attachments         <- getTicketAttachments ticketInfo
-        mapM (inspectAttachment ticketInfo) attachments
-
-    postTicketComment   <- asksZendeskLayer zlPostTicketComment
-    mapM_ postTicketComment (concat zendeskResponses)
+    _                   <- mapM (processTicket . ticketId) sortedTicketIds
 
     putTextLn "All the tickets has been processed."
 
--- a/src/Lib.hs
+++ b/src/Lib.hs
@@ -9,6 +9,8 @@ module Lib
     , processTickets
     , fetchTickets
     , showStatistics
+
+    , processBatchTickets
     ) where
 
 import           Universum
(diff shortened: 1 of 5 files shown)


src/Lib.hs
```diff
@@ -70,65 +73,85 @@ collectEmails = do
...
+
 
 processTickets :: App ()
 processTickets = do
-    cfg <- ask
-    sortedTicketIds <- processBatchTickets cfg
-    mapM_ processTicketAndId sortedTicketIds
+    sortedTicketIds     <- listAndSortTickets
+
+    zendeskResponses    <- forM sortedTicketIds $ \ticketInfo -> do
```
HirotoShioi: I think this will cause an issue.
This function first analyzes all the tickets the Zendesk, then start posting comments. Analysis will take a long time so if the process was killed (or failed), it will have to do everything all over again.

test/Spec.hs
```diff
@@ -1,2 +1,159 @@
...
+
+import           Universum
+
+import           Test.Hspec
+import           Test.Hspec.QuickCheck
+import           Test.QuickCheck
+import           Test.QuickCheck.Monadic
+
+import           Zendesk
+import           Lib
```
HirotoShioi: Can you make these imports explicit?

src/Lib.hs
ksaric: If a crash occurs, it's not going to be good in any case. In other cases, this is ok. But the plan was to switch to `processTicket` anyway, so I'll do that.

src/Zendesk/Types.hs
```diff
@@ -177,6 +224,29 @@ data TicketInfo = TicketInfo
     , ticketStatus :: !Text        -- ^ The status of the ticket
     } deriving (Eq, Show)
 
+
+-- TODO(ks): Newtype!
+instance Arbitrary TicketURL where
+    arbitrary = do
+        protocol    <- elements ["http://"]
```
HirotoShioi: `protocol` only has 1 element so we could just write it as `let protocol = "http://"` or we can add `https://` as one of the choices! so it would be `elements ["http://", "https://"]`

src/Zendesk/Types.hs
```diff
@@ -177,6 +224,29 @@ data TicketInfo = TicketInfo
...
+        ticketUrl    <- arbitrary
+        ticketTags   <- pure []
+        ticketStatus <- pure "open"
+
+        pure TicketInfo
+            { ticketId      = ticketId
+            , ticketUrl     = ticketUrl
+            , ticketTags    = ticketTags
+            , ticketStatus  = ticketStatus
+            }
```
HirotoShioi: This can be written much simpler like this `pure TicketInfo{..}`

src/Zendesk/Types.hs
ksaric: Yes, that is an option I left. We will fix this once we start wrapping it in a newtype.

src/Zendesk/Types.hs
ksaric: Which is not explicit. While I will switch from one style to the other, depending on the number of parameters, this is the correct way to do it and to be as explicit as possible.

PR #15, opened 2018-06-04
Co 294 - Basic statistic query completed

robcohen:
https://iohk.myjetbrains.com/youtrack/issue/TSD-17

 to run
```:main stats2 in ghci```
 
## Basic statistic query completed:
 - Will list # of open/closed tickets
 - Will print out each attachment and display size

## Todo:
- Migration of functionality to Statistics.hs
- Refactoring, code should be terse and less App ()

### DIFF
--- a/src/CLI.hs
+++ b/src/CLI.hs
@@ -17,7 +17,6 @@ data CLI
     | FetchTickets      -- ^ Fetch all the tickets in Zendesk
     | RawRequest String -- ^ Raw request to the given url
     | ShowStatistics    -- ^ Show statistics
-    | ShowStats2        -- ^ Show stats2
     deriving (Show)
 
 -- | Parser for ProcessTicket
@@ -47,8 +46,6 @@ cli = hsubparser $ mconcat
         (progDesc "Raw request to the given url"))
     , command "show-stats" (info (pure ShowStatistics)
         (progDesc "Print list of ticket Ids that agent has been assigned"))
-    , command "stats2" (info (pure ShowStats2)
-        (progDesc "Print more stats"))
     ]
 
 -- | Get CLI arguments from command line
--- a/src/CLI.hs
+++ b/src/CLI.hs
@@ -17,6 +17,7 @@ data CLI
     | FetchTickets      -- ^ Fetch all the tickets in Zendesk
     | RawRequest String -- ^ Raw request to the given url
     | ShowStatistics    -- ^ Show statistics
+    | ShowStats2        -- ^ Show stats2
     deriving (Show)
 
 -- | Parser for ProcessTicket
@@ -46,6 +47,8 @@ cli = hsubparser $ mconcat
         (progDesc "Raw request to the given url"))
     , command "show-stats" (info (pure ShowStatistics)
         (progDesc "Print list of ticket Ids that agent has been assigned"))
+    , command "stats2" (info (pure ShowStats2)
+        (progDesc "Print more stats"))
     ]
 
 -- | Get CLI arguments from command line
--- a/src/Statistics.hs
+++ b/src/Statistics.hs
@@ -1,0 +1,8 @@
+module CLI
+    ( CLI(..)
+    , getCliArgs
+    ) where
+
+import           Universum
+
+
--- a/src/Statistics.hs
+++ b/src/Statistics.hs
@@ -1,8 +1,26 @@
-module CLI
-    ( CLI(..)
-    , getCliArgs
-    ) where
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Statistics
+    ( statsExample ) where
 
 import           Universum
+import           Types (Attachment (..), Comment (..), Ticket (..), TicketId, TicketInfo (..),
+                        TicketList (..), TicketTag (..), ZendeskResponse (..), parseAgentId,
+                        parseComments, parseTickets, renderTicketStatus)
 
+-- | Filepath to token file
+tokenPath :: FilePath
+tokenPath = "./tmp-secrets/token"
 
+-- | Filepath to assign_to file
+assignToPath :: FilePath
+assignToPath = "./tmp-secrets/assign_to"
+
+statsExample :: IO ()
+statsExample = do
+  token <- (readFile tokenPath)  -- Zendesk token
+  assignFile <- readFile assignToPath  -- Select assignee
+  putTextLn "end of stats"
--- a/src/Statistics.hs
+++ b/src/Statistics.hs
@@ -4,23 +4,13 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Statistics
-    ( statsExample ) where
+    (filterTicketsByStatus) where
 
 import           Universum
 import           Types (Attachment (..), Comment (..), Ticket (..), TicketId, TicketInfo (..),
                         TicketList (..), TicketTag (..), ZendeskResponse (..), parseAgentId,
                         parseComments, parseTickets, renderTicketStatus)
 
--- | Filepath to token file
-tokenPath :: FilePath
-tokenPath = "./tmp-secrets/token"
+filterTicketsByStatus :: [TicketInfo] -> Text -> [TicketInfo]
+filterTicketsByStatus tickets status =  filter ((== status) . ticketStatus) tickets
 
--- | Filepath to assign_to file
-assignToPath :: FilePath
-assignToPath = "./tmp-secrets/assign_to"
-
-statsExample :: IO ()
-statsExample = do
-  token <- (readFile tokenPath)  -- Zendesk token
-  assignFile <- readFile assignToPath  -- Select assignee
-  putTextLn "end of stats"
--- a/src/Zendesk.hs
+++ b/src/Zendesk.hs
@@ -19,7 +19,7 @@ import           Network.HTTP.Simple (Request, addRequestHeader, getResponseBody
                                       setRequestMethod, setRequestPath)
 
 import           CLI (CLI (..), getCliArgs)
-import           Statistics
+import           Statistics (filterTicketsByStatus)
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/CLI.hs [hunk 1]
```diff
@@ -17,6 +17,7 @@ data CLI
     | FetchTickets      -- ^ Fetch all the tickets in Zendesk
     | RawRequest String -- ^ Raw request to the given url
     | ShowStatistics    -- ^ Show statistics
+    | ShowStats2        -- ^ Show stats2
```
ksaric: You can reuse the original `ShowStatistics`, no need for a different one.

src/CLI.hs
```diff
@@ -46,6 +47,8 @@ cli = hsubparser $ mconcat
         (progDesc "Raw request to the given url"))
     , command "show-stats" (info (pure ShowStatistics)
         (progDesc "Print list of ticket Ids that agent has been assigned"))
+    , command "stats2" (info (pure ShowStats2)
```
ksaric: Again, reuse the original `ShowStatistics`.

src/Statistics.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Statistics
```
ksaric: This is the file where all the functionality should go in. I understand you might have done this by example, but the whole point of creating this file is to have all related functionality in it. Move all the statistics code here, otherwise, we will have a ton of code conflicts.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
             printWarning
             tickets <- runApp (listTickets Assigned) cfg
             printTicketCountMessage tickets (cfgEmail cfg)
+        ShowStats2 -> do
+            putTextLn $ "Testing - Statistics associated with: "
+               <> cfgEmail cfg
+            tickets <- runApp (listTickets Assigned) cfg
+            statsExample
+            -- How many tickets are open
+            putTextLn $ "There are " <> show (length $ filter ((== "open") . ticketStatus) tickets) <> " open tickets."
```
ksaric: You can put things like `length $ filter ((== "open") . ticketStatus) tickets` in separate functions to increase readability.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+            -- How many tickets are closed
+            putTextLn $ "There are " <> show (length $ filter ((== "closed") . ticketStatus) tickets) <> " closed tickets."
```
ksaric: Same as above.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+            -- How many open tickets have attachments
+            let openTickets = filter (\ticket -> ticketStatus ticket == "closed") tickets --TODO: CHANGE TO OPEN (Only closed shows results as there are no open tickets according to the analyzer)
```
ksaric: It's a matter of preference, but I prefer to have anonymous functions transformed into actual functions with parameters.
`\ticket -> ticketStatus ticket == "closed"` == `isTicketClosed ticket = ticketStatus ticket == "closed"`.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+            runApp (showTicketsWithAttachments openTickets) cfg
+            putTextLn $ "End of statistics."
+
+showTicketsWithAttachments :: [TicketInfo] -> App ()
+showTicketsWithAttachments tickets = do
+  liftIO $ putTextLn "Scanning tickets for attachments"
```
ksaric: Follow indentation rules (4 spaces) and style guide used here - https://github.com/serokell/serokell-util/blob/master/serokell-style.md

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+  let checkTicketForAttachments :: TicketInfo -> App Bool
+      checkTicketForAttachments ticket = do
+        comments <- getTicketComments (ticketId ticket)
+        commentsWithAttachments <- pure $ (filter (\comment -> length (cAttachments comment) > 0) comments)
```
ksaric: Hm, hm. This seems too complicated.
When you find yourself doing things like these `var <- pure/return` it's usually an indication that you have to think harder about the issue. Why? Because `<-` "removes" the monadic context, while `pure/return` introduces it. So it kind of cancels out, see?
For example, try removing the function `checkTicketForAttachments` and doing everything, step by step. After you complete it, _then_ try packing things into functions. 
Abstractions (mostly) aren't invented, they are found, so you must take care not to invent things, but to gradually massage them into a correct form. 
Take my word for it, it's extremely hard to hit the correct abstraction the first time around.
So you simply don't try to do that.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+        pure $ not (null commentsWithAttachments)
+  ticketsWithAttachments <- filterM (\ticket -> checkTicketForAttachments ticket) tickets
```
ksaric: Again, feels like forcing things into monadic context.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+  liftIO $ putTextLn $ "There are " <> show (length ticketsWithAttachments) <> " open tickets with attachments"
+  showTicketsAttachmentInfo ticketsWithAttachments
+
+showTicketsAttachmentInfo :: [TicketInfo] -> App ()
+showTicketsAttachmentInfo tickets = do
+
+  let showTicketAttachments :: TicketInfo -> App ()
```
ksaric: If you are going to use this many `let` bindings, I suggest using `where`.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+              pure attachments'
+
+        let showTicketAttachmentInfo :: TicketInfo -> App ()
+            showTicketAttachmentInfo ticket'' = do
+              putText "Ticket #: "
+              print (ticketId ticket'')
+              comments <- getComments ticket''
+              putText "  Attachments: "
+              -- Attachment Name
+              mapM_ (\comment -> do
```
ksaric: Again, in here the anonymous functions make this hard to read.

src/Zendesk.hs
```diff
@@ -144,6 +145,69 @@ runZendeskMain = do
...
+                                  putText "  "
+                                  print $ aSize attachment
+                              ) (getAttachments comment)
+                    ) comments
+              pure ()
+
+        showTicketAttachmentInfo ticket
+
+  mapM_ showTicketAttachments tickets
+  pure ()
```
ksaric: This is not good. Functions return things, procedures return nothing. We want to return things since we want to be able to test functions. You will see in time, in the meantime, just return something you think is the result of this function.
What would that be? If you, for example, don't output anything in this function, would you be able to create a structure that you can return from this function and then output that function somewhere else?
Once you start thinking like that, you will start separating data and its effects, which is the key principle to understand for composable design.
You can see more here (it's Scala, but it's very similar) - https://www.youtube.com/watch?v=1gZAqJA2pEk

src/Zendesk.hs
```diff
@@ -141,9 +142,48 @@ runZendeskMain = do
...
+showTicketAttachments :: TicketInfo -> App ()
+showTicketAttachments ticket = do
+    putText "Ticket #: "
+    (putTextLn . show) (ticketId ticket)
+    comments <- getTicketComments (ticketId ticket)
+    mapM_ showCommentAttachments (comments)
+
+filterTicketsWithAttachments :: [TicketInfo] -> App [TicketInfo]
+filterTicketsWithAttachments tickets = do
+    let checkTicketForAttachments :: TicketInfo -> App Bool
```
ksaric: Let me give you an example.
Let's say you have:
```
filterTickets :: TicketInfo -> App Bool
filterTickets ticket = do
    comments <- getTicketComments (ticketId ticket) -- You get comments from a ticket using a monadic context
    pure . null $ comments
```
You start thinking "how can I remove this monadic context and push it out?". Can also be rewritten as:
```
filterTicketByComments :: TicketInfo -> App Bool
filterTicketByComments ticketInfo = do
    comments <- getTicketComments (ticketId ticket) -- You get comments from a ticket using a monadic context
    pure $ filterTickets comments

filterTickets :: [Comment] -> Bool
filterTickets comments = null comments
```
This seems kind of strange but it really isn't. The function `filterTickets` should do what the name says - filter tickets. Not fetch the ticket comments, filter that, and return it.
This buys you composability in the long run.

ksaric: I guess the example here is the fact you can extract the "impure" part out.

robcohen: Ok, reviewing now.

robcohen: isn't filterTickets lacking useful description? To me filterTickets should take something to filter with, not simply test if an array is empty or not. In this case shouldn't it be called something like ticketHasComments or checkTicketForComments?

ksaric: Yes, that's better. I was actually trying to show you a different thing, but you are correct, this is something I wrote down quickly. Let me fix the example.

robcohen: Ok. I will try to port over my work to your recent updates in develop.

ksaric: When I fixed the naming, the example got even more confusing. Let me try with an unrelated and obvious example:
```
printTicket :: TicketInfo -> App ()
printTicket ticketInfo = putStrLn . show $ ticketInfo
```
Can be:
```
showTicket :: TicketInfo -> String
showTicket ticketInfo = show ticketInfo

printTicket :: TicketInfo -> App ()
printTicket ticketInfo = putStrLn $ showTicket ticketInfo
```
See, `showTicket` is now pure. Anyway, hope it help, otherwise ignore my ramblings.

robcohen: So the issue I had is that you need to traverse all the comments of a ticket, then all the attachments of a comment, so to seperate things you'd need [tiicketinfo], [[Comments]], [[[Attachments]]] then try to count them properly (which is prone to error). I'm going to play with the refactoring more and see what's possible.

PR #17, opened 2018-06-12
[TSD-40] Send the message in to all requesters with no log in the ticket

HirotoShioi:
## Description
This PR will introduce responses to tickets with no logs attached.

## Related issue
https://iohk.myjetbrains.com/youtrack/issue/TSD-40


HirotoShioi:
Currently, I don't have a template response for "No log files". Asking carl to provide them!

src/Lib.hs
```diff
@@ -75,24 +75,24 @@ collectEmails = do
 
 
 processTicket :: TicketId -> App [ZendeskResponse]
-processTicket ticketId = do
+processTicket tid = do
```
ksaric: `tid` = `tId`

src/Lib.hs
```diff
@@ -231,6 +217,12 @@ getAttachmentsFromComment comments = do
     isAttachmentZip :: Attachment -> Bool
     isAttachmentZip attachment = "application/zip" == aContentType attachment
 
+-- | Get zendesk responses
+getZendeskResponses :: [Comment] -> [Attachment] -> TicketInfo -> App [ZendeskResponse]
```
ksaric: Isn't the mapping one `TicketInfo` - one `ZendeskResponse`? We should fix this, as noted above.

src/Lib.hs
```diff
@@ -310,3 +316,6 @@ filterAnalyzedTickets ticketsInfo =
     isTicketBlacklisted :: TicketInfo -> Bool
     isTicketBlacklisted TicketInfo{..} = ticketId `notElem` [9377,10815]
 
+    isGoguenTestnetTicket :: TicketInfo -> Bool
```
ksaric: `isTicketInGoguenTestnet`

src/Lib.hs
```diff
@@ -291,14 +283,28 @@ inspectAttachment ticketInfo@TicketInfo{..} att = do
...
 -- | Filter analyzed tickets
 filterAnalyzedTickets :: [TicketInfo] -> [TicketInfo]
 filterAnalyzedTickets ticketsInfo =
     filter ticketsFilter ticketsInfo
   where
     ticketsFilter :: TicketInfo -> Bool
-    ticketsFilter ticketInfo =
-        isTicketAnalyzed ticketInfo && isTicketOpen ticketInfo && isTicketBlacklisted ticketInfo
+    ticketsFilter ticketInfo = 
+        all (\predicate -> predicate ticketInfo) [ isTicketAnalyzed
```
ksaric: A complicated way to say a simple thing. Remember that you will have to maintain this code and that doing something smart may not be the best things to do - do something simple instead.
I know it seems kind of harsh to say that, but it's important that you understand how little changes like these make the code more complex. I actually have to think about this, which is something I like to avoid as much as possible.
And if this builds, isn't `all ticketInfo [ ... ]`  an option?

test/Spec.hs
```diff
@@ -150,7 +152,65 @@ processTicketSpec =
...
+                    let stubbedConfig :: Config
+                        stubbedConfig = withStubbedIOAndZendeskLayer stubbedZendeskLayer
+
+                    let appExecution :: IO [ZendeskResponse]
+                        appExecution = runApp (processTicket . ticketId $ ticketInfo) stubbedConfig
+
+                    zendeskResponses <- run appExecution
+                    assert $ not (null zendeskResponses) && not (isTaggedWithNoLogs zendeskResponses)
+
+isTaggedWithNoLogs :: [ZendeskResponse] -> Bool
```
ksaric: Can you insert the param name? I know this is also a convention, but I **have to** look at the type to know what this is doing, which is not helping readability.
Try to be as explicit as possible, it helps readability and maintainability.

src/Lib.hs
```diff
@@ -231,6 +217,12 @@ getAttachmentsFromComment comments = do
     isAttachmentZip :: Attachment -> Bool
     isAttachmentZip attachment = "application/zip" == aContentType attachment
 
+-- | Get zendesk responses
+getZendeskResponses :: [Comment] -> [Attachment] -> TicketInfo -> App [ZendeskResponse]
+getZendeskResponses comments attachments ticketInfo
+    | not (null attachments) = mapM (inspectAttachment ticketInfo) attachments
+    | not (null comments)    = pure <$> responseNoLogs ticketInfo
```
ksaric: `pure <$> responseNoLogs ticketInfo` == `responseNoLogs ticketInfo`

src/Lib.hs
```diff
@@ -75,24 +75,24 @@ collectEmails = do
 
 
 processTicket :: TicketId -> App [ZendeskResponse]
```
ksaric: We need to fix this. We should take the last comment attachments and process the last attachment.

test/Spec.hs
```diff
@@ -150,7 +152,65 @@ processTicketSpec =
...
+
+                    let stubbedConfig :: Config
+                        stubbedConfig = withStubbedIOAndZendeskLayer stubbedZendeskLayer
+
+                    let appExecution :: IO [ZendeskResponse]
+                        appExecution = runApp (processTicket . ticketId $ ticketInfo) stubbedConfig
+
+                    zendeskResponses <- run appExecution
+
+                    assert $ not (null zendeskResponses) && isTaggedWithNoLogs zendeskResponses
```
ksaric: Very nice :+1:

test/Spec.hs
```diff
@@ -150,7 +152,65 @@ processTicketSpec =
...
+
+        it "processes ticket, with attachments" $
+            forAll (listOf1 arbitrary) $ \(comments:: [Comment]) ->
+                forAll arbitrary $ \ticketInfo ->
+                    monadicIO $ do
+
+                    pre $ any (not . null . cAttachments) comments
```
ksaric: `all (not . null . cAttachments) comments` - we require that all the comments have attachments, right?

src/Lib.hs
HirotoShioi: My first impression is that fixing this could be a little tricky. I'll try to fix it today and if I couldn't get it done by today, then I'll have to ask for your help

src/Lib.hs
```diff
@@ -231,6 +215,22 @@ getAttachmentsFromComment comments = do
...
+getZendeskResponses comments attachments ticketInfo
+    | not (null attachments) = inspectAttachments ticketInfo attachments
+    | not (null comments)    = Just <$> responseNoLogs ticketInfo
+    | otherwise              = return Nothing
+
+-- | Inspect only the latest attchment
+inspectAttachments :: TicketInfo -> [Attachment] -> App (Maybe ZendeskResponse)
+inspectAttachments ticketInfo attachments = do
+    -- Inspecting the last element since it'll be the latest attachment
+    let latestAttachment = last <$> nonEmpty attachments
```
ksaric: How about `inspectAttachment ticketInfo =<< safeHead . sort $ attachments` where the `Ord` instance should be `AttachmentId`? Or the expanded form of the same which is even better:
```
lastAttachment <- safeHead . sort $ attachments
inspectAttachment ticketInfo lastAttachment
```

src/Lib.hs
```diff
@@ -231,6 +215,22 @@ getAttachmentsFromComment comments = do
...
+    case latestAttachment of
```
ksaric: This should be a bind - `latestAttachment >>= \att -> inspectAttachment ticketInfo att` or similar.
See how they align? `Nothing` goes to `Nothing` while `Just` goes to `Just`. 
You do work when it's `Just` otherwise return nothing.

test/Spec.hs
ksaric: Yes, good thinking but I didn't even notice we were returning multiple `ZendeskResponse`, my bad.

PR #18, opened 2018-06-12
[TSD-39] Caching using a simple SQLite database

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-39

A read-only implementation of the SQLite DB, still missing the write from web API side, but since this is already a large PR, will fix in another one.


test/Spec.hs
```diff
@@ -134,18 +134,18 @@ processTicketSpec =
...
-                                    { zlListTickets         = \_     -> pure listTickets
-                                    , zlGetTicketInfo       = \_     -> pure ticketInfo
-                                    , zlPostTicketComment   = \_     -> pure ()
-                                    , zlGetTicketComments   = \_     -> pure comments
-                                    , zlGetAttachment       = \_     -> pure mempty
+                                    { zlListAssignedTickets     = \_     -> pure listTickets
+                                    , zlGetTicketInfo           = \_     -> pure $ Just ticketInfo
+                                    , zlPostTicketComment       = \_     -> pure ()
+                                    , zlGetTicketComments       = \_     -> pure comments
+                                    , zlGetAttachment           = \_     -> pure mempty
```
HirotoShioi: Tests are failing here.
```
    • No instance for (Monoid AttachmentContent)
        arising from a use of ‘mempty’
```

src/DataSource/DB.hs
```diff
@@ -0,0 +1,162 @@
...
+    , zlGetTicketComments       = liftIO . getTicketComments
+    , zlPostTicketComment       = error "We won't use this for now!"
+    }
+
+-- Database instances
+-- FROM
+
+instance FromField TicketId where
+    fromField (Field (SQLInteger tId) _)    = Ok . TicketId . fromIntegral $ tId
+    fromField f                             = returnError ConversionFailed f "need a integer, ticket id"
```
HirotoShioi: Grammar error. "need `an` integer, ticket id" .  Other error message has same issues as well

HirotoShioi:
Also, I'm not really sure how we're going to utilize user datatype but I guess that'll be clarified in the next PR

HirotoShioi:
We should add `*.db` (or create db folder then ignore that file?) to the `.gitignore` so we don't accidentally upload `prod.db` to github by mistake.

ksaric:
*.db will be created in the next PR along with the git ignore rule. The rest is ok.

src/DataSource/DB.hs
```diff
@@ -0,0 +1,162 @@
...
+import           Database.SQLite.Simple.Ok
+import           Database.SQLite.Simple.ToField
+
+import           DataSource.Types (Attachment (..), AttachmentContent (..), AttachmentId (..),
+                                   Comment (..), CommentBody (..), CommentId (..), Config,
+                                   TicketId (..), TicketInfo (..), TicketStatus (..),
+                                   TicketTags (..), TicketURL (..), UserId (..), ZendeskLayer (..))
+
+
+-- https://ocharles.org.uk/blog/posts/2014-08-07-postgresql-simple-generic-sop.html
```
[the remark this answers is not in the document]
ksaric: Yes, that is why I pasted it here.

src/DataSource/DB.hs
ksaric: Interesting, I remember I read somewhere they can be interchanged, but it doesn't seem to be so - https://www.quickanddirtytips.com/education/grammar/a-versus-an

PR #19, opened 2018-06-15
[TSD-41] Run the debugger on everything pending, open, new, on-hold and unassigned for all agents.

HirotoShioi:
## Summary
This PR introduces few things
- Run debugger on pending, open, new, hold status
- Fetch agents using [Zendesk groups api](https://developer.zendesk.com/rest_api/docs/core/groups)
- Process all the tickets assigned to the agents

## Related issues
https://iohk.myjetbrains.com/youtrack/issue/TSD-41

## Note
This PR will require you to have `group` file with groupId written down in the `tmp_secrets` folder. I'll upload the file on the slack channel


HirotoShioi:
@ksaric Sorry I'm not sure which part of the code you're talking about.. Perhaps this one? https://github.com/input-output-hk/log-classifier/blob/hiroto/TSD-41/src/DataSource/Http.hs#L118

src/CLI.hs
```diff
@@ -14,9 +14,10 @@ import           Paths_log_classifier (version)
 -- use direct function calls.
 data CLI
     = CollectEmails     -- ^ Collect email addresses
+    | FetchAgents        -- ^ List agents
```
ksaric: Space :-)

src/DataSource/Http.hs
```diff
@@ -34,6 +35,7 @@ defaultConfig =
         , cfgZendesk            = "https://iohk.zendesk.com"
         , cfgToken              = ""
         , cfgEmail              = "daedalus-bug-reports@iohk.io"
+        , cfgGroupId            = GroupId 0
```
ksaric: Remove this. We don't require this much flexibility for now. We just need the _admin_ group for now.

src/DataSource/Http.hs
```diff
@@ -111,6 +115,14 @@ listAssignedTickets userId = do
 
     iterateTicketPages req
 
+listAgents :: forall m. (MonadIO m, MonadReader Config m) => GroupId -> m [User]
```
ksaric: `listAdminAgents` and remove groupId.

src/DataSource/Types.hs
```diff
@@ -127,13 +133,18 @@ tokenPath = "./tmp-secrets/token"
 assignToPath :: FilePath
 assignToPath = "./tmp-secrets/assign_to"
 
+-- | Filepath to group file
+groupPath :: FilePath
+groupPath = "./tmp-secrets/group"
```
ksaric: We need to remove this, not introduce more of it...

src/DataSource/Http.hs
```diff
@@ -132,6 +144,26 @@ iterateTicketPages req = do
         Just nextUrl -> liftIO $ go page0 nextUrl
         Nothing      -> pure page0
 
+iterateUserPages
```
ksaric: If this is copy and paste, can you instead just genetalize the signature to somethin more general?

src/Lib.hs
```diff
@@ -136,18 +157,20 @@ listAndSortTickets = do
...
+        agentIds = uId <$> agents
     -- We first fetch the function from the configuration
     listTickets <- asksZendeskLayer zlListAssignedTickets
     printText   <- asksIOLayer iolPrintText
 
-    printText $ "Classifier is going to process tickets assign to: " <> email
+    printText "Classifier is going to process tickets assigned to agents"
 
-    tickets     <- listTickets userId
+    ticketInfos     <- concat <$> mapM listTickets agentIds
```
ksaric: `traverse`?

src/Lib.hs
```diff
@@ -294,8 +317,11 @@ filterAnalyzedTickets ticketsInfo =
     isTicketAnalyzed TicketInfo{..} = (renderTicketStatus AnalyzedByScriptV1_0) `notElem` (getTicketTags tiTags)
     -- ^ This is showing that something is wrong...
 
+    unsolvedTicketStatus :: [TicketStatus]
+    unsolvedTicketStatus = TicketStatus <$> ["new", "open", "hold", "pending"]
```
ksaric: Use `map` when lifting in a list context, makes it more explicit and easier to read...

src/DataSource/Http.hs
```diff
@@ -145,7 +177,7 @@ postTicketComment ZendeskResponse{..} = do
     let req2 = addJsonBody
                    (Ticket
                        (Comment (CommentId 0) (CommentBody $ "**Log classifier**\n\n" <> zrComment) [] zrIsPublic (cfgAgentId cfg))
-                       (cfgAssignTo cfg)
+                       Nothing
```
ksaric: What if an _assignee_ is already selected? This will just write down a comment with no "owner"?

HirotoShioi: I've tested it, the classifier won't touch the assignee field at all. It just leave the field as it is.

src/DataSource/Http.hs
HirotoShioi: I've tried but it's a bit difficult for me to do it.. I'll tackle again tomorrow morning.

src/DataSource/Http.hs
```diff
@@ -66,6 +68,7 @@ emptyZendeskLayer = ZendeskLayer
     { zlGetTicketInfo           = \_     -> error "Not implemented zlGetTicketInfo!"
     , zlListRequestedTickets    = \_     -> error "Not implemented zlListRequestedTickets!"
     , zlListAssignedTickets     = \_     -> error "Not implemented zlListAssignedTickets!"
+    , zlListAdminAgents         =           error "Not implemented zlListAgents" -- This returns error during testing
```
HirotoShioi: Any test that uses ZendeskLayer fails because of this. Do you know any workaround for this?

src/DataSource/Types.hs
```diff
@@ -172,6 +176,8 @@ data ZendeskAPIUrl
     deriving (Eq, Generic)
 
 showURL :: ZendeskAPIUrl -> Text
+showURL AgentGroupURL                       = "/groups/41693007/users.json" -- magic number..?
+-- users.json?role[]=admin&role[]=agent returns InvalidUrl
```
HirotoShioi: I've tried using `users.json?role[]=admin&role[]=agent` to fetch the tickets it returns `Invalid Url` so I opted to use `/api/v2/users/{user_id}/groups.json` but in that case, we'll now have magic number.

src/DataSource/Http.hs
ksaric: Awesome, thanks for the confirmation.

src/DataSource/Http.hs
ksaric: What about something like:
```
iteratePages
    :: forall m a. (MonadIO m, MonadReader Config m, FromJSON a)
    => Request -> m [a]
iteratePages req = do ...
```
Use `parseJSON` instead of `parseUsers` and it's counterpart, and create a general structure that contains the page and results:
```
data PageResultList a = PageResultList
    { tlResults :: ![a]
    , nextPage  :: Maybe Text -- Text ?
    }
```
Try it and if you get stuck, ping me.

src/DataSource/Http.hs
ksaric: Ah, yes. It's something I lost a ton of time on - for now, you can just do `pure []` and be done with it, otherwise the `error` get's evaluated immediately - there is no parameter for it, see?

src/DataSource/Types.hs
ksaric: Hm. Well, you can use the `GroupId` and use a parameter, which would sort out the issue above and make testing easier, but don't include it in any configuration, send it as a magic number from some high-level module like `Lib`.

HirotoShioi: Okay! I've kind of solved this.

src/DataSource/Http.hs
HirotoShioi: Done! But I had to add Monad constraint to the emptyZendeskLayer (could be applicative)

HirotoShioi:
@ksaric 
Okay! Fixed 1 and 2! 
Regarding 3, is there an way to add some kind of equality constraint on `a`  such that `a` have to be either `User` or `TicketInfo`? Since type signature of `iteragePage` is `iteratePages :: forall m a. (MonadIO m, MonadReader Config m, FromJSON a)`, `a` could be anything but we only want `a` as either `User` or `TicketInfo`.

src/DataSource/Http.hs
HirotoShioi: Yea I think I'll add comment about them as well

src/DataSource/Http.hs
```diff
@@ -109,25 +109,32 @@ listAssignedTickets userId = do
...
+listAdminAgents :: forall m. (MonadIO m, MonadReader Config m) => m [User]
+listAdminAgents = do
+    cfg <- ask
+    let url = showURL AgentGroupURL
+    let req = apiRequest cfg url
 
+    iteratePages req
+
+-- | Iterate all the ticket pages and combine into a result.
+iteratePages :: forall m a. (MonadIO m, MonadReader Config m, FromJSON a)
```
HirotoShioi: Perhaps it would be better if we can add some kind of equality constraint on `a` that `a` have to be either `User` or `TicketInfo` but I'm not sure how you can do this.

src/DataSource/Types.hs
```diff
@@ -454,6 +436,29 @@ instance FromJSON Attachment where
...
+            , cPublic      = commentIsPublic
+            , cAuthor      = commentAuthorId
+            }
+
+-- Is there an way to make this better?
+instance (FromJSON a) => FromJSON (PageResultList a) where
+    parseJSON = withObject "ticketList" $ \o ->
+            PageResultList
+                <$> (o .: "tickets" <|>  o .: "users")
+                <*> o .: "next_page"
```
HirotoShioi: This does work but I feel like this is naive way to implement it.. Any better idea?

PR #21, opened 2018-06-22
Run debugger 1.1

HirotoShioi:
## Summary
- Update to V1.1
- Change email content
- Add tag `analyzed-by-script-v1.1`
- Update test suites

### DIFF
--- a/log-classifier.cabal
+++ b/log-classifier.cabal
@@ -1,5 +1,5 @@
 name:                log-classifier
-version:             1.0.0.0
+version:             1.1.0.0
 synopsis:            Log classifier for a Cardano node
 description:         Please see the README
 homepage:            https://github.com/input-output-hk/log-classifier#readme
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -290,6 +290,7 @@ data TicketInfo = TicketInfo
 data TicketTag
     = AnalyzedByScript      -- ^ Ticket has been analyzed
     | AnalyzedByScriptV1_0  -- ^ Ticket has been analyzed by the version 1.0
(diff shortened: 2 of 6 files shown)


log-classifier.cabal
```diff
@@ -1,5 +1,5 @@
 name:                log-classifier
-version:             1.0.0.0
+version:             1.1.0.0
```
ksaric: Nice catch.

src/DataSource/Types.hs
```diff
@@ -530,10 +529,9 @@ instance ToJSON CommentOuter where
                 ]
 
 instance ToJSON Ticket where
-    toJSON (Ticket comment assignee tags) =
+    toJSON (Ticket comment tags) =
         object  [ "ticket" .= object
                     [ "comment"     .= comment
-                    , "assignee_id" .= assignee
```
ksaric: This means we don't change the `assignee` ever, right?

test/Spec.hs
```diff
@@ -275,18 +275,23 @@ filterAnalyzedTicketsSpec =
                 \(ticketInfos :: [TicketInfo]) ->
                     length (filterAnalyzedTickets ticketInfos) `shouldBe` 0
 
-        it "should filter analyzed tickets" $
+        it "should filter analyzed tickets v1.0" $
             forAll (listOf $ genTicketWithFilteredTags ["analyzed-by-script-v1.0"]) $
                 \(ticketInfos :: [TicketInfo]) ->
                     length (filterAnalyzedTickets ticketInfos) `shouldBe` 0
 
+        it "should filter analyzed tickets v1.1" $
```
ksaric: Nice, even though the other scenario should be tested when testing - the one where all of the tickets have the tag, or some of the issues have the tag. Never mind for now.

PR #22, opened 2018-06-26
[TSD-57] Fix the run for unassigned tickets

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-57

Repeat the run from https://iohk.myjetbrains.com/youtrack/issue/TSD-41 for unassigned tickets.

### DIFF
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -47,8 +47,8 @@ module DataSource.Types
 
 import           Universum
 
-import           Data.Aeson (FromJSON, ToJSON, Value, object, parseJSON, toJSON, withObject, (.:),
-                             (.=))
+import           Data.Aeson (FromJSON, ToJSON, Value (Object), object, parseJSON, toJSON,
+                             withObject, (.:), (.=))
 import           Data.Aeson.Types (Parser)
 import           Data.Text (pack)
 import           LogAnalysis.Types (Knowledge)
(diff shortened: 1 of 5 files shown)


src/Lib.hs
```diff
@@ -119,16 +119,24 @@ processTicket tId = do
...
 processTickets :: App ()
 processTickets = do
-    sortedTicketIds     <- listAndSortTickets
-    _                   <- mapM (processTicket . tiId) sortedTicketIds
+    sortedTicketIds             <- listAndSortTickets
+    sortedUnassignedTicketIds   <- listAndSortUnassignedTickets
+
+    let allTickets = sortedTicketIds <> sortedUnassignedTicketIds
+
+    _                   <- mapM (processTicket . tiId) sortedUnassignedTicketIds
```
ksaric: This is for the current run.

src/DataSource/Types.hs
```diff
@@ -177,6 +179,7 @@ showURL :: ZendeskAPIUrl -> Text
 showURL AgentGroupURL                       = "users.json?role%5B%5D=admin&role%5B%5D=agent"
 showURL (UserRequestedTicketsURL userId)    = "/users/" <> toURL userId <> "/tickets/requested.json"
 showURL (UserAssignedTicketsURL userId)     = "/users/" <> toURL userId <> "/tickets/assigned.json"
+showURL (UserUnassignedTicketsURL)          = "/search.json?query=type%3Aticket%20assignee%3Anone&sort_by=created_at&sort_order=asc"
```
HirotoShioi: We could use `urlEncode` function from  `http-types` to encode this.
https://hackage.haskell.org/package/http-types-0.12.1/docs/Network-HTTP-Types.html#v:urlEncode

So the query would look something like this:
`"/search.json?query=" <> decodeUtf8 (urlEncode False someQuery)`

I couldn't find url encoding function that returns `Text`, if there's one, that'll be better..

src/DataSource/Types.hs
```diff
@@ -457,11 +460,27 @@ class FromPageResultList a where
     fromPageResult :: Value -> Parser (PageResultList a)
 
 instance FromPageResultList TicketInfo where
-    fromPageResult = withObject "ticketList" $ \o ->
+    fromPageResult obj = asum
+        [ ticketListParser obj
+        , resultsTicketsParser obj
+        ]
+      where
```
HirotoShioi: I was very interested in how you can implement this in a clean way. Thank you!

src/Lib.hs
HirotoShioi: Just curious, can we fetch all the ticket with single fetching function (basically we fetch all the "new", "open", "pending", "hold" tickets and we don't care about whether they're assigned or not) or is that not really a good idea?

ksaric: I don't see why we _have to_ do that. We can if the future requirements of the project dictate so.

src/DataSource/Types.hs
ksaric: Yeah, I wouldn't say it's practical to import a library for a single function. In this case, it's cleaner to write our own.

src/DataSource/Types.hs
ksaric: :+1:

PR #23, opened 2018-06-27
[TSD-58] Add logging feature

HirotoShioi:
https://iohk.myjetbrains.com/youtrack/issue/TSD-57

- Add logging feature to the classifier so we can later on analyze the run result.
- Replace `[Text]` with `TicketTags` where possible.
- Save the log result on a log file
- The classifer will create "logs" folder in order to store it.

### DIFF
--- a/src/DataSource/Http.hs
+++ b/src/DataSource/Http.hs
@@ -23,9 +23,9 @@ import           DataSource.Types (Attachment (..), AttachmentContent (..), Comm
                                    CommentBody (..), CommentId (..), Config (..),
                                    FromPageResultList (..), IOLayer (..), PageResultList (..),
                                    Ticket (..), TicketId (..), TicketInfo (..), TicketTag (..),
-                                   User, UserId (..), ZendeskAPIUrl (..), ZendeskLayer (..),
-                                   ZendeskResponse (..), parseComments, parseTicket,
-                                   renderTicketStatus, showURL)
+                                   TicketTags (..), User, UserId (..), ZendeskAPIUrl (..),
+                                   ZendeskLayer (..), ZendeskResponse (..), parseComments,
+                                   parseTicket, renderTicketStatus, showURL)
 
 
 -- | The default configuration.
@@ -164,6 +164,7 @@ postTicketComment
 postTicketComment ZendeskResponse{..} = do
     cfg <- ask
 
+    let ticketTags = TicketTags (renderTicketStatus AnalyzedByScriptV1_1 : getTicketTags zrTags)
     let url  = showURL $ TicketsURL zrTicketId
     let req1 = apiRequest cfg url
     let req2 = addJsonBody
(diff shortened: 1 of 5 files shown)


src/Lib.hs
```diff
@@ -95,27 +95,37 @@ fetchAgents = do
...
+    whenJust zendeskResponse $ \response -> do
+        printZendeskResponse response
+        postTicketComment response
 
     pure zendeskResponse
 
+printZendeskResponse :: ZendeskResponse -> App ()
+printZendeskResponse zendeskResponse = do
+    let ticketTags = zrTags zendeskResponse
+    printText <- asksIOLayer iolPrintText
```
ksaric: Move this out of this function to keep it pure. Call it `printZendeskResponse` and fetch `iolPrintText` and print it in the function above.

src/DataSource/Http.hs
```diff
@@ -150,6 +150,7 @@ postTicketComment
 postTicketComment ZendeskResponse{..} = do
     cfg <- ask
 
+    let ticketTags = TicketTags (renderTicketStatus AnalyzedByScriptV1_1 : getTicketTags zrTags)
```
ksaric: Shouldn't this work and not override the tags?

src/Lib.hs
HirotoShioi: Nice point!

src/DataSource/Http.hs
HirotoShioi: Well I guess how Zendesk API works is that if we post list of tags to the ticket, it'll override the existing ones..

HirotoShioi: Additional notes.
Okay when I fetch the analysis result, I do see some of the tags not being overwritten, which is good but strange.. I'll look into this today.

src/Lib.hs
```diff
@@ -95,27 +97,39 @@ fetchAgents = do
...
 
-    printText "Process finished, please see the following url"
-    printText $ "https://iohk.zendesk.com/agent/tickets/" <> show (getTicketId tId)
+    postTicketComment   <- asksZendeskLayer zlPostTicketComment
+    
+    whenJust zendeskResponse $ \response -> do
+        let formattedTags = formatZendeskResponseTags response
+        printText formattedTags
+        appendF <- asksIOLayer iolAppendFile
+        appendF "logs/analysis-result.log" (show (getTicketId tId) <> " " <> formattedTags <> "\n")
```
ksaric: I don't like this `append` since it could introduce issues due to file changes, it could be easier to add it all in a String/Text/ByteString and add it into the file at the end of the function. These mutations over time always end up in some weird situations...

There are other issues connected with that, like losing log info due to program crashing or large overhead in memory (really negligible), but I guess it's more obvious/transparent to the user if there is an issue in the application, there should be one in the logs as well.

Other than that, kudos for updating our IO layer. A bonus question would be - how would you test `append` or the function? Does `append` reflect the function behaviour? :thinking: 
Seems like there is something there :wink:

PR #24, opened 2018-06-27
[TSD-59] Keep existing tags, fields as it is after processing the ticket

HirotoShioi:
https://iohk.myjetbrains.com/youtrack/issue/TSD-59

Description:
For some reason, Zendesk changes the values of the custom field to `null` when posting the comment unless we explicitly tell Zendesk not to change it.
Also, we've been overwriting the existing tags.

This PR will fix these issues.

### DIFF
--- a/src/DataSource/DB.hs
+++ b/src/DataSource/DB.hs
@@ -23,6 +23,7 @@ import           Database.SQLite.Simple.ToField (ToField (..))
 
 import           DataSource.Types (Attachment (..), AttachmentContent (..), AttachmentId (..),
                                    Comment (..), CommentBody (..), CommentId (..), Config,
+                                   TicketField (..), TicketFieldId (..), TicketFieldValue (..),
                                    TicketId (..), TicketInfo (..), TicketStatus (..),
                                    TicketTags (..), TicketURL (..), UserId (..), ZendeskLayer (..))
 
@@ -46,6 +47,17 @@ cachedZendeskLayer = ZendeskLayer
 -- Database instances
 -- FROM
 
+instance FromField TicketFieldId where
+    fromField (Field (SQLInteger tfId) _)     = Ok . TicketFieldId . fromIntegral $ tfId
+    fromField f                             = returnError ConversionFailed f "need a text, ticket status"
+
+instance FromField TicketFieldValue where
+    fromField (Field (SQLText tfValue) _)     = Ok . TicketFieldValue $ tfValue
+    fromField f                             = returnError ConversionFailed f "need a text, ticket status"
+
+instance FromField [TicketField] where
+    fromField _ = undefined
+
 instance FromField TicketId where
     fromField (Field (SQLInteger tId) _)    = Ok . TicketId . fromIntegral $ tId
     fromField f                             = returnError ConversionFailed f "need an integer, ticket id"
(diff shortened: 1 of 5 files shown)


HirotoShioi:
I've tested and it works.
I've tried adding other fields (`C.C`, `Type`) to see if the classifier removes/changes it but it didn't.
If we were to make changes to the ticket information, we should be careful about it..!

src/DataSource/DB.hs
```diff
@@ -46,6 +47,17 @@ cachedZendeskLayer = ZendeskLayer
...
+instance FromField TicketFieldId where
+    fromField (Field (SQLInteger tfId) _)     = Ok . TicketFieldId . fromIntegral $ tfId
+    fromField f                             = returnError ConversionFailed f "need a text, ticket status"
+
+instance FromField TicketFieldValue where
+    fromField (Field (SQLText tfValue) _)     = Ok . TicketFieldValue $ tfValue
+    fromField f                             = returnError ConversionFailed f "need a text, ticket status"
+
+instance FromField [TicketField] where
+    fromField _ = undefined
```
ksaric: ?? There should be a warning, try fixing this.

src/DataSource/Http.hs
```diff
@@ -159,24 +162,33 @@ iteratePages req = do
...
-                   )
-                   req1
-    void $ liftIO $ apiCall (pure . encodeToLazyText) req2
+    let responseTicket = createResponseTicket (cfgAgentId cfg) ticketInfo zendeskResponse
+    let url  = showURL $ TicketsURL (zrTicketId zendeskResponse)
+    let req = addJsonBody responseTicket (apiRequest cfg url)
+    void $ liftIO $ apiCall (pure . encodeToLazyText) req
+
+-- | Create response ticket
+createResponseTicket :: Integer -> TicketInfo -> ZendeskResponse -> Ticket
```
ksaric: Nice.

src/DataSource/Types.hs
```diff
@@ -261,9 +265,9 @@ data RequestType
 data ZendeskResponse = ZendeskResponse
     { zrTicketId :: !TicketId
     , zrComment  :: !Text
-    , zrTags     :: ![Text] -- TODO(ks): This should be wrapped
+    , zrTags     :: !TicketTags
```
ksaric: It's really regarding the previous PR, but good :+1:

src/DataSource/Types.hs
```diff
@@ -428,6 +463,8 @@ instance Arbitrary TicketInfo where
             , tiUrl         = ticketUrl
             , tiTags        = ticketTags
             , tiStatus      = ticketStatus
+            , tiField       = ticketField
+            , tiCustomField  = ticketCustomField
```
ksaric: Space :telescope:

src/DataSource/Types.hs
```diff
@@ -294,12 +298,27 @@ newtype CommentOuter = CommentOuter {
 
 -- | Zendesk ticket
 data Ticket = Ticket
-    { tComment :: !Comment
+    { tComment     :: !Comment
     -- ^ Ticket comment
-    , tTag     :: ![Text]
+    , tTag         :: !TicketTags
+    , tField       :: ![TicketField]
```
ksaric: I'm not entirely convinced that we need this. Let's have a chat first.

src/DataSource/DB.hs
HirotoShioi: Yes I will! Thank you!

src/DataSource/Types.hs
HirotoShioi: Sure!

HirotoShioi:
@ksaric Thank you for checking!
I've resolved merge conflict, fix the format of the log.
If there's not problem, I can merge it!

PR #25, opened 2018-07-04
[TSD-47] Create SQLite schema and insert data using HTTP

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-47

Run it for almost a day and it skips "phantom" tickets and inserts the data into the database. 
After it's in the database, the data is available almost instantly.
There is still plenty of work to be done, one of the priorities is to add exception handling because it's getting very painful to run something just to see it crash a couple of minutes after that.

### DIFF
--- a/src/DataSource/Http.hs
+++ b/src/DataSource/Http.hs
@@ -11,7 +11,9 @@ module DataSource.Http
 
 import           Universum
 
+import           Control.Concurrent (threadDelay)
 import           Control.Monad.Reader (ask)
+
 import           Data.Aeson (FromJSON, ToJSON, Value, encode, parseJSON)
 import           Data.Aeson.Text (encodeToLazyText)
 import           Data.Aeson.Types (Parser, parseEither)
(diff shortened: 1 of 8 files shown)


src/DataSource/Http.hs
```diff
@@ -37,7 +40,7 @@ defaultConfig =
         { cfgAgentId            = 0
         , cfgZendesk            = "https://iohk.zendesk.com"
         , cfgToken              = ""
-        , cfgEmail              = "daedalus-bug-reports@iohk.io"
+        , cfgEmail              = "kristijan.saric@iohk.io"
```
HirotoShioi: You should revert this to `daedalus-bug-reports@iohk.io`. This address is used when the classifier is posting comments on Zendesk.

src/DataSource/Http.hs
```diff
@@ -130,35 +153,93 @@ listUnassignedTickets = do
...
     iteratePages req
 
+-- | Export tickets from Zendesk - https://developer.zendesk.com/rest_api/docs/core/incremental_export
+-- NOTE: If count is less than 1000, then stop paginating.
+-- Otherwise, use the next_page URL to get the next page of results.
+getExportedTickets
+    :: forall m. (MonadIO m, MonadReader Config m)
+    => ExportFromTime
+    -> m [TicketInfo]
+getExportedTickets time = do
```
HirotoShioi: This function seems to be printing a lot of unnecessary printing.. Can you clean it up?

test/Spec.hs
```diff
@@ -315,6 +314,12 @@ validShowURLSpec =
                 let typedURL    = showURL $ TicketCommentsURL ticketId
                     untypedURL  = "/tickets/" <> show (getTicketId ticketId) <> "/comments.json"
                 in  typedURL == untypedURL
+        it "returns valid ExportDataByTimestamp" $ do
```
HirotoShioi: This test is failing

src/DataSource/Http.hs
ksaric: We need to move some of these to CLI, this doesn't make sense here.

src/DataSource/Http.hs
ksaric: I wouldn't say it's "unnecessary" since I required that in order to run it and find three major issues.
I have an alternative though, will do it in a separate ticket.

test/Spec.hs
ksaric: Fixed. Speaking of failing tests, did you add these:
-      processes ticket, with comments FAILED [1]
-      processes ticket, with no attachments FAILED [2]
-      processes ticket, with attachments FAILED [3]

src/DataSource/Http.hs
HirotoShioi: Thanks!

test/Spec.hs
HirotoShioi: Yes I did, and it's strange because those tests are passing in the develop branch.

PR #26, opened 2018-07-09
[TSD-70] Exception handling when reading log files

HirotoShioi:
This PR will introduce exception handling on reading log files so that the program will not crash due to the corrupted log file.

- Fix test cases (Don't use external file when testing)
- Exception handling on reading log files. (Example tciket id: 15066)

https://iohk.myjetbrains.com/youtrack/issue/TSD-70

### DIFF
--- a/log-classifier.cabal
+++ b/log-classifier.cabal
@@ -48,6 +48,7 @@ library
                      , optparse-applicative
                      , reflection
                      , regex-tdfa
+                     , safe-exceptions
                      , text
                      , time
                      , universum
--- a/log-classifier.cabal
+++ b/log-classifier.cabal
@@ -17,6 +17,7 @@ library
   hs-source-dirs:      src
   exposed-modules:     Classify
                        CLI
+                       Exceptions
                        Lib
                        LogAnalysis.Classifier
                        LogAnalysis.KnowledgeCSVParser
--- a/log-classifier.cabal
+++ b/log-classifier.cabal
@@ -20,6 +20,7 @@ library
                        Exceptions
                        Lib
                        LogAnalysis.Classifier
+                       LogAnalysis.Exceptions
                        LogAnalysis.KnowledgeCSVParser
                        LogAnalysis.Types
                        Regex
--- a/log-classifier.cabal
+++ b/log-classifier.cabal
@@ -49,7 +49,6 @@ library
                      , optparse-applicative
                      , reflection
                      , regex-tdfa
-                     , safe-exceptions
                      , text
                      , time
                      , universum
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -11,7 +11,6 @@ module DataSource.Types
     , CommentId (..)
     , CommentBody (..)
     , CommentOuter (..)
-    , LogFiles(..)
     , PageResultList (..)
     , RequestType (..)
     , DeletedTicket (..)
@@ -348,10 +347,6 @@ data Comment = Comment
 -- | Outer comment ??
 newtype CommentOuter = CommentOuter {
       coComment :: Comment
-    }
-
-newtype LogFiles = LogFiles {
-    getLogFiles :: [LByteString]
     }
 
 -- | Zendesk ticket
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -11,6 +11,7 @@ module DataSource.Types
     , CommentId (..)
     , CommentBody (..)
     , CommentOuter (..)
+    , LogFiles(..)
     , PageResultList (..)
     , RequestType (..)
     , DeletedTicket (..)
@@ -347,6 +348,10 @@ data Comment = Comment
 -- | Outer comment ??
 newtype CommentOuter = CommentOuter {
       coComment :: Comment
+    }
+
+newtype LogFiles = LogFiles {
+    getLogFiles :: [LByteString]
     }
 
 -- | Zendesk ticket
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -411,7 +411,6 @@ data TicketTag
     | AnalyzedByScriptV1_2  -- ^ Ticket has been analyzed by the version 1.2
     | NoKnownIssue          -- ^ Ticket had no known issue
     | NoLogAttached         -- ^ Log file not attached
-    | CannotReadZipFile
 
 newtype UserId = UserId
     { getUserId :: Int
@@ -812,5 +811,4 @@ renderTicketStatus AnalyzedByScriptV1_0 = "analyzed-by-script-v1.0"
 renderTicketStatus AnalyzedByScriptV1_1 = "analyzed-by-script-v1.1"
 renderTicketStatus AnalyzedByScriptV1_2 = "analyzed-by-script-v1.2"
 renderTicketStatus NoKnownIssue         = "no-known-issues"
-renderTicketStatus CannotReadZipFile    = "cannot-read-zip-file"
 renderTicketStatus NoLogAttached        = "no-log-files"
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -80,6 +80,8 @@ newtype App a = App { runAppBase :: ReaderT Config IO a }
              , MonadIO
              , MonadBase IO
              , MonadUnliftIO
+             , MonadCatch
+             , MonadThrow
              )
 
 instance MonadBaseControl IO App where
@@ -409,6 +411,7 @@ data TicketTag
     | AnalyzedByScriptV1_2  -- ^ Ticket has been analyzed by the version 1.2
     | NoKnownIssue          -- ^ Ticket had no known issue
     | NoLogAttached         -- ^ Log file not attached
+    | CannotReadZipFile
 
 newtype UserId = UserId
     { getUserId :: Int
@@ -809,4 +812,5 @@ renderTicketStatus AnalyzedByScriptV1_0 = "analyzed-by-script-v1.0"
 renderTicketStatus AnalyzedByScriptV1_1 = "analyzed-by-script-v1.1"
 renderTicketStatus AnalyzedByScriptV1_2 = "analyzed-by-script-v1.2"
 renderTicketStatus NoKnownIssue         = "no-known-issues"
+renderTicketStatus CannotReadZipFile    = "cannot-read-zip-file"
 renderTicketStatus NoLogAttached        = "no-log-files"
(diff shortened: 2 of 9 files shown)


src/Lib.hs
```diff
@@ -335,21 +335,30 @@ inspectAttachments ticketInfo attachments = runMaybeT $ do
...
+                , zrTags        = TicketTags [renderTicketStatus CannotReadZipFile]
+                , zrIsPublic    = (cfgIsCommentPublic config)
+                }
+        )
 
 
 -- | Given number of file of inspect, knowledgebase and attachment,
 -- analyze the logs and return the results.
-inspectAttachment :: Config -> TicketInfo -> AttachmentContent -> ZendeskResponse
+inspectAttachment :: (MonadCatch m) => Config -> TicketInfo -> AttachmentContent -> m ZendeskResponse
```
ksaric: Don't touch this, this is all pure, no need to infect that part of the code with impurities.

src/Lib.hs
```diff
@@ -335,21 +335,30 @@ inspectAttachments ticketInfo attachments = runMaybeT $ do
     lastAttachment  <- MaybeT . pure $ lastAttach
     att             <- MaybeT $ getAttachment lastAttachment
 
-    pure $ inspectAttachment config ticketInfo att
+    catchAny (inspectAttachment config ticketInfo att)
```
ksaric: We are not looking for `Any` exception, we need specific exceptions we can throw and catch. Like here - https://github.com/input-output-hk/cardano-sl/blob/feature/CBR-163/wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs

HirotoShioi: Thanks for the advice! Will look into it!

src/Lib.hs
HirotoShioi: Got it!

HirotoShioi:
Implemented obvious ones. Now I'm going to try to solve exception `*** Exception: Codec.Compression.Zlib: premature end of compressed data stream`

src/Lib.hs
```diff
@@ -353,51 +355,54 @@ inspectAttachments ticketInfo attachments = runMaybeT $ do
...
+          where
+            renderErrorTag :: ClassifierExceptions -> TicketTags
+            renderErrorTag ReadZipFileException   = TicketTags [renderErrorCode SentLogCorrupted]
+            renderErrorTag DecompressionException = TicketTags [renderErrorCode DecompressionFailure]
 
 -- | Given number of file of inspect, knowledgebase and attachment,
 -- analyze the logs and return the results.
-inspectAttachment :: Config -> TicketInfo -> AttachmentContent -> ZendeskResponse
-inspectAttachment Config{..} ticketInfo@TicketInfo{..} attContent = do
+inspectAttachment :: Config -> TicketInfo -> [LByteString]-> ZendeskResponse
```
ksaric: Why did you strip the newtype wrapper?

src/DataSource/Types.hs
```diff
@@ -77,6 +77,8 @@ newtype App a = App { runAppBase :: ReaderT Config IO a }
              , MonadReader Config
              , MonadIO
              , MonadBase IO
+             , MonadCatch
```
ksaric: I really think we don't need this, unless you want to re-export, but that's a whole different story.

log-classifier.cabal
```diff
@@ -46,6 +47,7 @@ library
                      , optparse-applicative
                      , reflection
                      , regex-tdfa
+                     , safe-exceptions
```
ksaric: Are you sure we need this? http://hackage.haskell.org/package/universum-1.2.0/docs/Universum-Exception.html

src/Lib.hs
```diff
@@ -205,7 +206,8 @@ processTicket tId = do
     let ticketInfo      = fromMaybe (error "No ticket info") mTicketInfo
     zendeskResponse     <- getZendeskResponses comments attachments ticketInfo
     postTicketComment   <- asksZendeskLayer zlPostTicketComment
-
+   
+    -- What should we do if the zendeskResponse was Nothing?
```
ksaric: The point of this refactoring is to add _error details_ and remove `Maybe` where possible. Maybe you would have an idea what to do with it if you had specific exceptions you could catch here?

src/Util.hs
```diff
@@ -20,10 +23,18 @@ extractLogsFromZip numberOfFiles file = do
     mTake n = Map.fromDistinctAscList . take n . Map.toAscList
 
 -- | Read zipe file
-readZip :: LByteString -> Either Text (Map FilePath LByteString)
+-- toArchiveOrFail is a partial function, we need to use tryDeep to catch the exception and throw it 
+-- upwards
+-- Why tryDeep instead of try? It's because we need to fully evaluate bytestrings in order to catch
+-- the decompression issue.
+readZip :: (MonadCatch m, MonadIO m) => LByteString -> m (Map FilePath LByteString)
 readZip rawzip = case Zip.toArchiveOrFail rawzip of
```
ksaric: Did you try it out? Does it catch the exceptions?

log-classifier.cabal
HirotoShioi: I've tried importing `tryDeep` from `Universum.Exceptions` but didn't work. I'm not sure what they mean by `re-export`.
`cardano-sl` 's `new-wallet` component is importing `safe-exceptions` even though it's using `universum`
https://github.com/input-output-hk/cardano-sl/blob/develop/wallet-new/cardano-sl-wallet-new.cabal#L175

src/Util.hs
HirotoShioi: Yes, I did try to use `try` at first since it has less constraints but it didn't catch the decompression error.
I can try again.

src/Lib.hs
HirotoShioi: Ah okay, I'd tackle on that part once I'm done with handling the decompression issue..

log-classifier.cabal
ksaric: Ok, keep it there, just try using functions from `Universum` as much as possible.

src/DataSource/Types.hs
ksaric: Ah, sorry, I thought that was from the module exports :grin: 
Should have looked in more detail.

src/DataSource/Types.hs
```diff
@@ -347,6 +350,10 @@ newtype CommentOuter = CommentOuter {
       coComment :: Comment
     }
 
+newtype LogFiles = LogFiles {
```
ksaric: I prefer the formatting seen here - https://github.com/input-output-hk/log-classifier/pull/26/files#diff-ee661ebeb72d3f09577f0fd782917c5aR359 since it doesn't change any other lines besides the ones you change which is helpful.

test/Spec.hs
```diff
@@ -218,17 +194,100 @@ processTicketSpec =
...
+
+                        let appExecution :: IO ZendeskResponse
+                            appExecution = runApp (processTicket ticketId) stubbedConfig
+
+                        eZendeskResponse <- run (try appExecution)
+
+                        -- Check we have some comments.
+                        assert $ isLeft (eZendeskResponse :: Either ProcessTicketExceptions ZendeskResponse)
+                        whenLeft eZendeskResponse $ \processException ->
+                            assert $ processException == TicketInfoNotFound ticketId
```
ksaric: Nice.

src/Lib.hs
```diff
@@ -426,28 +437,51 @@ getAttachmentsFromComment comments = do
...
     getAttachment   <- asksZendeskLayer zlGetAttachment
 
-    let lastAttach :: Maybe Attachment
-        lastAttach = safeHead . reverse . sort $ attachments
-
-    lastAttachment  <- MaybeT . pure $ lastAttach
-    att             <- MaybeT $ getAttachment lastAttachment
-
-    pure $ inspectAttachment config ticketInfo att
+    mAtt <- runMaybeT $ do
```
ksaric: This seems overly verbose. Can you move it around a bit to make it a bit more concise?

test/Spec.hs
```diff
@@ -218,17 +194,99 @@ processTicketSpec =
...
+
+                    let appExecution :: IO ZendeskResponse
+                        appExecution = runApp (processTicket . tiId $ ticketInfo) stubbedConfig
+
+                    eZendeskResponse <- run (try appExecution)
+
+                    -- Check it throws exception
+                    -- How do I check if it returns proper exception?
+                    assert $ isLeft (eZendeskResponse :: Either ProcessTicketExceptions ZendeskResponse)
+                    whenLeft eZendeskResponse $ \processException ->
```
ksaric: Nice work.

src/Lib.hs
HirotoShioi: I'll try it out. Thank you.

src/Lib.hs
```diff
@@ -449,28 +460,29 @@ getAttachmentsFromComment comments = do
...
     getAttachment   <- asksZendeskLayer zlGetAttachment
 
-    let lastAttach :: Maybe Attachment
-        lastAttach = safeHead . reverse . sort $ attachments
-
-    lastAttachment  <- MaybeT . pure $ lastAttach
-    att             <- MaybeT $ getAttachment lastAttachment
-
-    pure $ inspectAttachment config ticketInfo att
+    lastAttach <- handleMaybe (safeHead . reverse . sort $ attachments)
```
ksaric: `handleMaybe . safeHead . reverse . sort $ attachments`

test/Spec.hs
```diff
@@ -167,13 +143,13 @@ processTicketSpec =
...
 
-                        let appExecution :: IO (Maybe ZendeskResponse)
+                        let appExecution :: IO ZendeskResponse
                             appExecution = runApp (processTicket . tiId $ ticketInfo) stubbedConfig
 
                         zendeskResponse <- run appExecution
 
                         -- Check we have some comments.
-                        assert $ isJust zendeskResponse
+                        assert $ (not . null . zrComment) zendeskResponse
```
ksaric: `assert . not . null . zrComment $ zendeskResponse`

src/Lib.hs
```diff
@@ -449,28 +460,29 @@ getAttachmentsFromComment comments = do
...
 getZendeskResponses comments attachments ticketInfo
     | not (null attachments) = inspectAttachments ticketInfo attachments
-    | not (null comments)    = Just <$> responseNoLogs ticketInfo
-    | otherwise              = return Nothing
+    | not (null comments)    = responseNoLogs ticketInfo
+    | otherwise              = throwM $ CommentAndAttachmentNotFound (tiId ticketInfo)
+    -- No attachment, no comments means something is wrong with ticket itself
 
 -- | Inspect only the latest attachment. We could propagate this
 -- @Maybe@ upwards or use an @Either@ which will go hand in hand
```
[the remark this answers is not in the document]
HirotoShioi: Good catch! I'll update it.

PR #28, opened 2018-07-10
[TSD-72]  Add option for running log-classifier on local zip

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-72

The added option of running the log-classifier locally on a local zip archive.


src/LogAnalysis/Types.hs
```diff
@@ -44,7 +45,16 @@ data Knowledge = Knowledge
     -- ^ Text describing how to solve the issue
     ,  kFAQNumber :: !Text
     -- ^ The FAQ number that will be displayed on the official Cardano FAQ page
-    } deriving (Show)
+    }
+
+instance Show Knowledge where
```
HirotoShioi: Good to know that you can actually do this.

src/Lib.hs
```diff
@@ -356,6 +358,37 @@ inspectAttachments ticketInfo attachments = runMaybeT $ do
...
+-- | Inspection of the local zip.
+-- This function prints out the analysis result on the console.
+inspectLocalZipAttachment :: FilePath -> App ()
+inspectLocalZipAttachment filePath = do
+
+    config          <- ask
+    printText       <- asksIOLayer iolPrintText
+
+    -- Read the zip file
+    fileContent     <- liftIO $ BS.readFile filePath
```
HirotoShioi: I think we should use `readFile` from ioLayer. If we try to implement test case on this function, it'lll try to read an external file which probably does not exist. Using ioLayer can prevent this.

ksaric: Yes, I know that - that's why I added these functions in the first place. However, this function is an `LByteString` while the "regular" one is a `ByteString`.
Since we don't want to introduce another function for lazy reading and the rest of the functions are pretty decently tested, this should be good.

HirotoShioi: Got it, thank you!

PR #30, opened 2018-07-12
Rob/ts21 install guide

robcohen:
[TSD-21 - Add instructions for building log-classifier](https://iohk.myjetbrains.com/youtrack/issue/TSD-21)

Instructions for Configuring and Building log-classifier with stack. The nix part of the guide and the scripts need to be worked on as a seperate task.


README.md
```diff
@@ -28,7 +31,7 @@ This is a use case diagram. Use case diagrams overview the usage requirements fo
 
 ## Requirements
 
-In order to use this, the user must be IOHK Zendesk agent and be able to provide email and password to the classifier.
+In order to use this, the user must be possess an IOHK Zendesk agent ID and API key.
```
mdimjasevic: There is a "be" left over. A correct way to write it would be: "In order to use this, the user must possess an IOHK Zendesk agent ID and API key."

robcohen: Good catch, thank you.

robcohen:
So that's from enabling nix. It was on while I was debugging some of the Nix scripts, I've turned it off. Things should compile now.

robcohen:
It doesn't need the change the default.nix, but we may as well include it as there is no Zendesk.hs anymore.

PR #32, opened 2018-07-16
[TSD-77] Fix two documentation issues that caused Hoogle to break

mdimjasevic:
https://iohk.myjetbrains.com/youtrack/issue/TSD-77

Hoogle cannot generate documentation for the log-classifier project due to incorrect documentation formatting in two places. These need to be fixed so that Hoogle can be used with the project.

### DIFF
--- a/src/Lib.hs
+++ b/src/Lib.hs
@@ -556,7 +556,7 @@ filterAnalyzedTickets ticketsInfo =
     unsolvedTicketStatus = map TicketStatus ["new", "open", "hold", "pending"]
 
     isTicketOpen :: TicketInfo -> Bool
-    isTicketOpen TicketInfo{..} = tiStatus `elem` unsolvedTicketStatus-- || ticketStatus == "new"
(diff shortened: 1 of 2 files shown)


src/Lib.hs
```diff
@@ -556,7 +556,7 @@ filterAnalyzedTickets ticketsInfo =
     unsolvedTicketStatus = map TicketStatus ["new", "open", "hold", "pending"]
 
     isTicketOpen :: TicketInfo -> Bool
-    isTicketOpen TicketInfo{..} = tiStatus `elem` unsolvedTicketStatus-- || ticketStatus == "new"
+    isTicketOpen TicketInfo{..} = tiStatus `elem` unsolvedTicketStatus-- ^ ticketStatus == "new"
```
ksaric: You can drop this comment, this was left and we don't need it anymore.

mdimjasevic:
I am not sure as to how to make sure that Hoogle can properly generate documentation. I tried with `rm -rf .stack-work/hoogle` and then  `stack hoogle`. Is that enough? @ksaric , your input would be appreciated.

ksaric:
@mdimjasevic Yes, as I already mentioned, that should be enough.

PR #33, opened 2018-07-18
[TSD-80] Nix on BuildKite CI Integration

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-80

Adding CI to the project.

### DIFF
--- a/CI.md
+++ b/CI.md
@@ -5,6 +5,7 @@ These are the instructions for generating nix expressions for buildkite CI for l
 ```
 $ cd $LOG_CLASSIFIER_ROOT
 $ cabal2nix . > default.nix
+$ nix-shell -p nix-prefetch-git
 nix-shell $ nix-prefetch-git https://github.com/NixOS/nixpkgs.git > nixpkgs.json
 ```
 
--- a/CI.md
+++ b/CI.md
@@ -1,13 +1,13 @@
 # Buildkite
-```
-$ nix-shell -p nix-prefetch-git
 
-$ nix-prefetch-git https://github.com/NixOS/nixpkgs.git > nixpkgs.json
- 
-```
+These are the instructions for generating nix expressions for buildkite CI for log-classifier.
 
-cabal2nix . > default.nix
+```
+$ cd $LOG_CLASSIFIER_ROOT
+$ cabal2nix . > default.nix
+nix-shell $ nix-prefetch-git https://github.com/NixOS/nixpkgs.git > nixpkgs.json
+```
 
 NOTE: 
   - NIX_PATH - Get from https://github.com/NixOS/nixpkgs, click clone, copy ZIP link, rename to .tar.gz.
-  - buildkite pipeline  is now: nix-build -I nixpkgs=$NIX_PATH release1.nix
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+[![Build status](https://badge.buildkite.com/50d1fa0b6ed53deb7bca3b81419fd2c65e7c22f67d2101e552.svg)](https://buildkite.com/input-output-hk/log-classifier)
```
ksaric: Nice.

mdimjasevic:
Just a naming clarification - is this TSD-49 as given in the PR's title or TSD-80 as given in @ksaric 's initial comment?

default.nix
```diff
@@ -1,12 +1,28 @@
...
-''
+{ mkDerivation, aeson, array, attoparsec, base, bytestring
+, containers, directory, generics-sop, hspec, http-conduit
+, monad-control, mtl, optparse-applicative, QuickCheck, reflection
+, regex-tdfa, resource-pool, sqlite-simple, stdenv, text, time
+, transformers-base, universum, zip-archive
+}:
+mkDerivation {
+  pname = "log-classifier";
+  version = "1.1.0.0";
```
ksaric: Version is `1.2.0.0` now.

shell.nix
```diff
@@ -1,18 +1,46 @@
...
+
+  f = { mkDerivation, aeson, array, attoparsec, base, bytestring
+      , containers, directory, generics-sop, hspec, http-conduit
+      , monad-control, mtl, optparse-applicative, QuickCheck, reflection
+      , regex-tdfa, resource-pool, sqlite-simple, stdenv, text, time
+      , transformers-base, universum, zip-archive
+      }:
+      mkDerivation {
+        pname = "log-classifier";
+        version = "1.1.0.0";
```
ksaric: Hm. This seems to be duplicated. Can't you import this since we have it already?

CI.md
```diff
@@ -0,0 +1,14 @@
+# Buildkite
+
+These are the instructions for generating nix expressions for buildkite CI for log-classifier.
+
+```
+$ cd $LOG_CLASSIFIER_ROOT
+$ cabal2nix . > default.nix
```
ksaric: I don't think that `cabal2nix` will capture all of the contents of the Stack project. Ideally, we should use `stack2nix` here, it might explain why you needed to add `Semigroup` - because it picked up the latest version of GHC, disregarding the version we put out in `stack.yaml`?
Anyway, good enough for now, you can just create another ticket for this specific issue.

ksaric:
@mdimjasevic Good catch, fixed.

mdimjasevic:
Can we please have a (short) description of what is achieved with this PR?

CI.md
robcohen: I tried to use stack2nix, but got errors like:
```error: Package ‘Win32-2.5.4.1’ in /home/coder/Nextcloud/IOHK/log-classifier/default.nix:1943 is not supported on ‘x86_64-unknown-linux-gnu’, refusing to evaluate.```

I spent some time trying to fix this but I couldn't crack it. Any idea how to fix this?

ksaric: No, but re-read my comment above and my comments in the channel.

PR #36, opened 2018-07-23
[TSD-71] Generalize HTTP functions to a layer.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-71

Moving all specific HTTP network functions to another layer.


ksaric:
@mdimjasevic They are needed here if we want the code to work as is. 
Having said that, I have to learn a lot about `RankNTypes` as well since it's not quite as simple a concept as I initially thought :grimacing:

PR #38, opened 2018-07-26
[TSD-86] Create visual architecture of the project

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-86

Adding some documentation and refactoring the code to match it.


ksaric:
Whoops. Good catch.
The CI is broken so I didn't even notice that.

PR #40, opened 2018-07-27
[TSD-87] Run the debugger on specific tag.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-87

Run the debugger on the specific tag, the tag is `to_be_analysed`.

### DIFF
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -165,7 +165,6 @@ asksDBLayer getter = do
     Config{..} <- ask
     pure $ getter cfgDBLayer
 
-
 -- TODO(ks): Move these three below to CLI!
 -- | Path to knowledgebase
 knowledgebasePath :: FilePath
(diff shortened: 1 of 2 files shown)


src/DataSource/Types.hs
```diff
@@ -847,5 +847,6 @@ renderTicketStatus AnalyzedByScript     = "analyzed-by-script"
 renderTicketStatus AnalyzedByScriptV1_0 = "analyzed-by-script-v1.0"
 renderTicketStatus AnalyzedByScriptV1_1 = "analyzed-by-script-v1.1"
 renderTicketStatus AnalyzedByScriptV1_2 = "analyzed-by-script-v1.2"
+renderTicketStatus ToBeAnalyzed         = "to_be_analysed" -- https://iohk.zendesk.com/agent/admin/tags
```
mdimjasevic: I can't access this link (https://iohk.zendesk.com/agent/admin/tags) so I can't see what's there. Anyhow, I thought `to-be-analyzed` could be instead of `to_be_analysed` to achieve consistency with other tags: they were written in American English and with dashes instead of underscored.

ksaric: The link can be seen by Zendesk admins (currently two of us from TSD).
Yes, I agree, but that's something we had no control over - it was done by the helpdesk team and it's what it is.

PR #45, opened 2018-08-01
[TSD-96] Fix getAttachmentsFromComment

HirotoShioi:
YouTrack: https://iohk.myjetbrains.com/youtrack/issue/TSD-96

Currently, the classifier skips any zip file with the content type of `application/x-zip-compress` (More details are here: https://iohk.myjetbrains.com/youtrack/issue/TSD-94#comment=93-26314). This PR will fix this issue.

### DIFF
--- a/src/DataSource/Types.hs
+++ b/src/DataSource/Types.hs
@@ -499,7 +499,8 @@ instance Arbitrary Attachment where
     arbitrary = Attachment
         <$> arbitrary
         <*> pure "http://attach.com"
-        <*> pure "application/zip"  -- TODO(ks): More random...
+        -- TODO(ks): More random...
+        <*> elements ["application/zip", "application/x-zip-compressed"]
         <*> arbitrary
 
 instance Arbitrary CommentBody where
--- a/src/Lib.hs
+++ b/src/Lib.hs
@@ -5,6 +5,7 @@ module Lib
     ( runZendeskMain
     , collectEmails
     , getZendeskResponses
+    , getAttachmentsFromComment
     , processTicket
     , processTicketSafe
     , processTickets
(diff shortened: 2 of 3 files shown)


src/DataSource/Types.hs
```diff
@@ -499,7 +499,8 @@ instance Arbitrary Attachment where
     arbitrary = Attachment
```
mdimjasevic: Is this test code? If so, why is it under `src/` and not `test/`? Can you move it?

ksaric: This is no test code. These are the instances for the `Arbitrary` typeclass. If you don't have any specific and good reason (like the _db_, for example, which I'm still not really sure about either), the instances of the specific type belong where the type is defined.
You can view this as a general arbitrary instance, while we can add more specific ones like we do in the test module(s).
The term you are looking to google is "orphan instances", that may answer some of your questions.

mdimjasevic: Thanks for the pointer. I'm new to Haskell so I'm still trying to understand this.

But if these arbitrary instances are not used anywhere in the source code, and only in the test code, shouldn't they be defined somewhere under the `test` directory?

HirotoShioi: From my perspective, it's easy to maintain the instances if we define in the single module. 

If we start saying "This typeclass is for this module only so we should defined it there"
1. It'll be much harder to maintain the instances since they're defined in separate modules. If we keep everything in `Types.hs` then I know where to look into if I have to update/define new instance.
2. There's definitely going be "gray zone" type classes being introduced. At first we only use that instance for specific module but then we realize we need it for another module as well which will bring confusion. I'm new to Haskell as well so I can't say for sure that which typeclasses are for specific use only or generic. If we maintain them in single file, I don't have to worry about which typeclasses are domain specific and which isn't.

Then there's also orphan instances as well.

PR #47, opened 2018-08-03
[TSD-93] Exception handling on DBLayer

HirotoShioi:
YouTrack: https://iohk.myjetbrains.com/youtrack/issue/TSD-93


src/DataSource/Exceptions.hs
```diff
@@ -0,0 +1,35 @@
+module DataSource.Exceptions where
```
ksaric: Move this to the `DB` module...

src/DataSource/DB.hs
```diff
@@ -147,7 +149,13 @@ connDBLayer = DBLayer
 
 -- | The connection pooled Zendesk layer. Used for database querying.
 -- We need to sync occasionaly.
-connPoolDataLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m) => DBConnPool -> DataLayer m
+connPoolDataLayer :: forall m. (MonadBaseControl IO m
```
ksaric: This formatting is wrong. Not sure if this is `stylish-haskell` or `hindent` or something else, but it should look like this:
```
connPoolDataLayer 
    :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m, MonadCatch m)
    => DBConnPool
    -> DataLayer m
```
Or if you want to split it into newlines (the max width should be 100, which is not really upheld in this project, we can fix that easily):
```
connPoolDataLayer 
    :: forall m. 
    ( MonadBaseControl IO m
    , MonadIO m
    , MonadReader Config m
    , MonadCatch m
    )
    => DBConnPool
    -> DataLayer m
```

src/DataSource/DB.hs
```diff
@@ -396,9 +410,9 @@ createSchema conn = liftIO $ execute_ conn
 \       FOREIGN KEY(`attachment_id`) REFERENCES comment_attachment(aId)     \
 \    )"
 
-insertTicketInfo :: forall m. (MonadIO m) => Connection -> TicketInfo -> m ()
+insertTicketInfo :: forall m. (MonadIO m, MonadCatch m) => Connection -> TicketInfo -> m ()
 insertTicketInfo conn TicketInfo{..} =
-    liftIO $ executeNamed conn "INSERT INTO ticket_info (tiId, tiRequesterId, tiAssigneeId, tiUrl, tiTags, tiStatus) \
+    checkError (InsertTicketInfoException tiId) $ executeNamed conn "INSERT INTO ticket_info (tiId, tiRequesterId, tiAssigneeId, tiUrl, tiTags, tiStatus) \
```
ksaric: This seems too heavy. The nice thing about functions is that you can compose them easily. 
Focus on making a single function do what it's supposed to do, not having a function that does everything for every case?
I'll be more specific after you try again.

src/DataSource/DB.hs
```diff
@@ -430,22 +444,31 @@ insertCommentAttachments conn Comment{..} Attachment{..} =
...
 
 -- | Delete all data.
-deleteAllData :: forall m. (MonadIO m) => Connection -> m ()
+deleteAllData :: forall m. (MonadIO m, MonadCatch m) => Connection -> m ()
 deleteAllData conn = do
     deleteCommentAttachments conn
     deleteTicketComments conn
     deleteTickets conn
 
+checkError :: (MonadIO m, MonadCatch m) => DBException -> IO () -> m ()
```
ksaric: IMHO, this is the function that kind of does too much.

log-classifier.cabal
```diff
@@ -30,6 +30,7 @@ library
                        Util
   other-modules:       Paths_log_classifier
                        DataSource.DB
+                       DataSource.Exceptions
```
ksaric: We don't need this, I think.

log-classifier.cabal
```diff
@@ -51,6 +52,7 @@ library
                      , optparse-applicative
                      , reflection
                      , regex-tdfa
+                     , safe-exceptions
```
ksaric, line 54: You really like this `safe-exceptions` library, don't you :smile: 
Fine, include it then.

src/DataSource/DB.hs
```diff
@@ -22,18 +22,20 @@ module DataSource.DB
...
 
+import           Control.Exception.Safe (Handler(..), catches)
 import           Control.Monad.Trans.Control (MonadBaseControl)
 
 import           Data.Pool (Pool, createPool, withResource)
 import           Data.Text (split)
 
 import           Database.SQLite.Simple (FromRow (..), NamedParam (..), SQLData (..), close,
-                                         executeNamed, execute_, field, open, queryNamed, query_)
+                                         executeNamed, execute_, field, open, queryNamed, query_, FormatError(..))
```
ksaric: This is just from my eyes :eye:, but it's not formatted?

ksaric (commented):
Left a couple of comments to take a look at.

src/DataSource/DB.hs
HirotoShioi: I was wondering how the style should be while coding. Thank you for telling me!

src/DataSource/DB.hs
HirotoShioi: Good catch! I'll fix it

src/DataSource/DB.hs
```diff
@@ -202,8 +209,8 @@ instance FromField UserId where
     fromField f                             = returnError ConversionFailed f "need an integer, user id"
 
 instance FromField TicketURL where
-    fromField (Field (SQLText tURL) _)      = Ok . TicketURL $ tURL
-    fromField f                             = returnError ConversionFailed f "need a text, ticket url"
+    fromField (Field (SQLText tURL) _) = Ok . TicketURL $ tURL
```
ksaric: I think you changed this a couple of times now. Leave it aligned with the others.

src/DataSource/DB.hs
```diff
@@ -163,7 +168,9 @@ connPoolDataLayer connPool = DataLayer
 
 
 -- | The connection pooled database layer. Used for database modification.
-connPoolDBLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m) => DBConnPool -> DBLayer m
+connPoolDBLayer :: forall m. ( MonadBaseControl IO m, MonadIO m, MonadReader Config m, MonadCatch m)
+                => DBConnPool
```
ksaric: 4 spaces.

src/DataSource/DB.hs
```diff
@@ -147,7 +150,9 @@ connDBLayer = DBLayer
 
 -- | The connection pooled Zendesk layer. Used for database querying.
 -- We need to sync occasionaly.
-connPoolDataLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m) => DBConnPool -> DataLayer m
+connPoolDataLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m, MonadCatch m)
```
ksaric: 4 spaces.

ksaric (commented):
Can you try adding some tests here? It might help clear out some of the code.

test/Spec.hs
```diff
@@ -631,4 +640,51 @@ getAttachmentsFromCommentSpec =
...
             zipFileContentTypes :: [Text]
-            zipFileContentTypes = ["application/zip", "application/x-zip-compressed"]
\ No newline at end of file
+            zipFileContentTypes = ["application/zip", "application/x-zip-compressed"]
+
+------------------------------------------------------------
+-- Tests for database operations
+------------------------------------------------------------
+
+deleteAllDataSpec :: Spec
```
ksaric: Good work. Will take a more detailed look when you think you covered the exceptional situations that can arise from the database.

HirotoShioi: Yes, I'll try to add more test cases. I think I can use mockery to create temporary database with certain conditions!

HirotoShioi:
@ksaric 
Today, I've tried to implement positive test (insert records and see if I can query it), but it seems like with our current implementation, not all the data are stored to the database (e.g. We're not storing `tiField`, `tiCustomField` from `TicketInfo`).

I can implement some test where I query part of the data (Like comment body, id). Do you think that'd be enough?

src/DataSource/DB.hs
```diff
@@ -358,47 +375,54 @@ getAttachmentContent conn Attachment{..} =
...
-\                                                                           \
-\    CREATE TABLE `attachment_content` (                                    \
-\       `attachment_id` INTEGER,                                            \
-\       `content`   BLOB NOT NULL,                                          \
-\       PRIMARY KEY(attachment_id),                                         \
-\       FOREIGN KEY(`attachment_id`) REFERENCES comment_attachment(aId)     \
-\    )"
-
-insertTicketInfo :: forall m. (MonadIO m) => Connection -> TicketInfo -> m ()
+createSchema conn = do
```
HirotoShioi, line 382: So the reason why I've made change is because all the database related tests fails with the message of `no such table` except for `ticket_info` table.
I'm not really sure why making this change solves the issue.
I'll look more into it,, perhaps reverting the changes.

test/Spec.hs
```diff
@@ -631,4 +642,145 @@ getAttachmentsFromCommentSpec =
...
+
+------------------------------------------------------------
+-- Tests for database operations
+------------------------------------------------------------
+
+-- Currently, this is how datas are stored in the DB, this can be changed in the future.
+type DBTicketInfo = (TicketId, UserId, Maybe UserId, TicketURL, TicketTags, TicketStatus)
+type DBTicketComment = (CommentId, TicketId, CommentBody, Bool, Integer)
+type DBCommentAttachment = (AttachmentId, CommentId, Text, Text, Int)
+type DBAttachmentContent = (Integer, ByteString)
```
HirotoShioi: This is due to the fact that the data that are stored in the database are slightly different from the original data.

HirotoShioi (commented):
I'll leave some comments here to clarify some of the changes that I've made.

src/DataSource/DB.hs
```diff
@@ -57,7 +67,7 @@ withDatabase dbName dbOperation =
         dbOperation
 
 -- | A production resource closing function.
-withProdDatabase :: forall m a. (MonadIO m) => (Connection -> IO a) -> m a
+withProdDatabase :: forall m a. (MonadIO m, MonadCatch m) => (Connection -> IO a) -> m a
```
ksaric: Wait. What do we catch here?

src/DataSource/DB.hs
ksaric: That's okay, this is even better. Maybe we should put this into a transaction to set it to be atomic, but since we use it once, should be fine as it is.

test/Spec.hs
ksaric: Seems like the test is telling us we need to add newtype wrappers.

test/Spec.hs
```diff
@@ -631,4 +642,145 @@ getAttachmentsFromCommentSpec =
...
+insertCommentAttachmentsSpec =
+    describe "insertCommentAttachmentsSpec" $ modifyMaxSuccess (const 200) $ do
+        prop "should insert comment and its attachment to the database" $
+            \(comment :: Comment) (attachment :: Attachment) ->
+                monadicIO $ do
+                    attachments <- run $
+                        withDBSchema ":memory:" (\conn -> do
+                            insertCommentAttachments conn comment attachment
+                            attachments <- queryNamed conn
+                                "SELECT * FROM comment_attachment WHERE aId = :id"
```
ksaric: The test is telling you that you need to add this query.

ksaric (changes requested):
A couple of smaller fixes.

src/DataSource/DB.hs
HirotoShioi: Good catch, this is not needed. There were some function with same issue so I've applied fixes to those as well.

test/Spec.hs
HirotoShioi: I've looked into it and it seems like I'm not using the proper query 
https://github.com/input-output-hk/log-classifier/blob/develop/src/DataSource/DB.hs#L351
(I'm supposed to query by comment id, not attachment id). I'll fix this.

HirotoShioi:
@ksaric 
```
type DBTicketComment = (CommentId, TicketId, CommentBody, Bool, Integer)
type DBCommentAttachment = (AttachmentId, CommentId, Text, Text, Int)
type DBAttachmentContent = (Integer, ByteString)
```
Seems like the test is telling us we need to add newtype wrappers.

Do you mean wrapping primitive type such as `Text`, `Integer`, `Int` with newtype?
If so, I suggest to create a new task for it to make this PR as minimal as possible.

ksaric:
> If so, I suggest to create a new task for it to make this PR as minimal as possible.

@HirotoShioi, In that case, let's not complicate this PR more, focus on finishing up the current state and let's merge it.

test/Spec.hs
```diff
@@ -632,3 +643,145 @@ getAttachmentsFromCommentSpec =
...
+type DBCommentAttachment = (AttachmentId, CommentId, Text, Text, Int)
+type DBAttachmentContent = (Integer, ByteString)
+
+deleteAllDataSpec :: Spec
+deleteAllDataSpec =
+    describe "deleteAllData" $ modifyMaxSuccess (const 200) $ do
+       prop "should delete all datas from database" $
+           monadicIO $ do
+               (tickets, ticketComments, commentAttachments, attachmentContents) <- 
+                   run $ withDBSchema ":memory:" (\conn -> do
```
ksaric: `run $ withDBSchema ":memory:" $ \conn -> do` or `run . withDBSchema ":memory:" $ \conn -> do`?

test/Spec.hs
```diff
@@ -632,3 +643,145 @@ getAttachmentsFromCommentSpec =
...
+                assert $ isLeft (eResult :: Either DBLayerException ())
+
+insertTicketCommentsSpec :: Spec
+insertTicketCommentsSpec =
+    describe "insertTicketComments" $ modifyMaxSuccess (const 200) $ do
+      prop "should insert ticketId and its comments to the dabatase" $
+          \(ticketId :: TicketId) (comment :: Comment) ->
+              monadicIO $ do
+               dbComments <- run $
+                    withDBSchema ":memory:" (\conn -> do
```
ksaric: Remove parens.

test/Spec.hs
```diff
@@ -632,3 +643,145 @@ getAttachmentsFromCommentSpec =
...
+              monadicIO $ do
+                eResult <- run . try $
+                    withConnection ":memory:" (\conn -> insertTicketComments conn ticketId comment)
+
+                assert $ isLeft (eResult :: Either DBLayerException ())
+
+-- | Perform an action to the database with schema
+withDBSchema :: String -> (Connection -> IO a) -> IO a
+withDBSchema dbpath action =
+    withConnection dbpath (\conn -> do
```
ksaric: Remove parens.

test/Spec.hs
```diff
@@ -632,3 +643,145 @@ getAttachmentsFromCommentSpec =
...
+                    assert $ isLeft (eResult :: Either DBLayerException ())
+
+insertTicketInfoSpec :: Spec
+insertTicketInfoSpec =
+    describe "insertTicketInfo" $ modifyMaxSuccess (const 200) $ do
+      prop "should insert ticket info to the dabatase" $
+          \(ticketInfo :: TicketInfo) ->
+              monadicIO $ do
+                dbTicketInfos <- run $
+                    withDBSchema ":memory:" (\conn -> do
```
ksaric: Remove parens.

src/DataSource/DB.hs
```diff
@@ -163,7 +175,9 @@ connPoolDataLayer connPool = DataLayer
 
 
 -- | The connection pooled database layer. Used for database modification.
-connPoolDBLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m) => DBConnPool -> DBLayer m
+connPoolDBLayer :: forall m. ( MonadBaseControl IO m, MonadIO m, MonadReader Config m, MonadCatch m)
```
ksaric: `:: forall ...` in a new line.

src/DataSource/DB.hs
```diff
@@ -147,7 +157,9 @@ connDBLayer = DBLayer
 
 -- | The connection pooled Zendesk layer. Used for database querying.
 -- We need to sync occasionaly.
-connPoolDataLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m) => DBConnPool -> DataLayer m
+connPoolDataLayer :: forall m. (MonadBaseControl IO m, MonadIO m, MonadReader Config m)
```
ksaric: `:: forall ...` in a new line.

ksaric (changes requested):
Minor fixes.

ksaric (approved):

PR #49, opened 2018-08-03
[TSD-91] Create changelog and changes documentation about the project.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-91

Created changelog and attached dates to it.


ksaric:
@HirotoShioi I don't know, I guess that depends on the amount of work. If there are a few changes, we can pack them with the updated version.

PR #50, opened 2018-08-09
[WIP] TSD-85 - Attempt #2 with Either and minimal

mdimjasevic:
This implements error handling with an `Either` in function return types to reflect the possibility of failure.


log-classifier.cabal
```diff
@@ -47,7 +47,7 @@ library
                      , monad-control
                      , transformers-base
                      , unliftio
-                     -- 
+                     --
```
ksaric: Generally, it's best to leave unchanged lines so they don't confuse people.

src/DataSource/Http.hs
```diff
@@ -16,6 +16,7 @@ import           Control.Monad.Reader (ask)
 import           Data.Aeson (parseJSON)
 import           Data.Aeson.Text (encodeToLazyText)
 import           Data.List (nub)
+import           Data.Text (pack)
```
ksaric: You don't need `pack`, you can look at https://github.com/serokell/universum/blob/master/src/Universum/String/Conversion.hs

src/HttpLayer.hs
```diff
@@ -40,41 +42,62 @@ emptyHTTPNetworkLayer = HTTPNetworkLayer
...
-    addRequestHeader "Content-Type" "application/json" $
-    setRequestBasicAuth
-        (encodeUtf8 cfgEmail <> "/token")
-        (encodeUtf8 cfgToken) $
-    parseRequest_ (toString (cfgZendesk <> path))
+apiRequest
+    :: Config
+    -> Text
+    -> Either String Request
+apiRequest Config{..} u = mapLeft show $ catches buildRequest handlerList
```
ksaric: I don't think we need an `Either` here. All our URLs are "internal" and will fail if incorrect anyway.
I would otherwise consider this, but it seems to be difficult for you right now.

src/HttpLayer.hs
```diff
@@ -40,41 +42,62 @@ emptyHTTPNetworkLayer = HTTPNetworkLayer
...
+    handlerHTTP = Handler $ \(ex :: HttpException) -> error $ show ex
     path :: Text
     path = "/api/v2" <> u
 
 -- | Api request but use absolute path
-apiRequestAbsolute :: Config -> Text -> Request
+apiRequestAbsolute
+    :: Config
+    -> Text
+    -> Either String Request
```
ksaric: Same here.

src/HttpLayer.hs
```diff
@@ -40,41 +42,62 @@ emptyHTTPNetworkLayer = HTTPNetworkLayer
...
 apiRequestAbsolute Config{..} u =
-    addRequestHeader "Content-Type" "application/json" $
+    Right <$> addRequestHeader "Content-Type" "application/json" $
     setRequestBasicAuth (encodeUtf8 cfgEmail <> "/token") (encodeUtf8 cfgToken) $
     parseRequest_(toString u)
 
 -- | Request PUT
-addJsonBody :: forall a. ToJSON a => a -> Request -> Request
-addJsonBody body req = setRequestBodyJSON body $ setRequestMethod "PUT" req
+addJsonBody :: forall a. ToJSON a => a -> Request -> Either String Request
```
ksaric: As a general rule, if you just wrap `Either` with a single constructor (`Right`), you probably don't need it.
Can you rewrite this to catch any exceptions and in that case to return it?

src/HttpLayer.hs
```diff
@@ -89,4 +112,8 @@ apiCallSafe parser req = do
     v <- getResponseBody <$> httpJSON req
     pure $ parseEither parser v
 
-
+-- | The mapLeft function takes a function and applies it to an Either value
+-- iff the value takes the form Left _.
+mapLeft :: (a -> c) -> Either a b -> Either c b
```
ksaric: You don't need this. You can use the function `either`.

src/HttpLayer.hs
```diff
@@ -40,41 +42,62 @@ emptyHTTPNetworkLayer = HTTPNetworkLayer
...
     -> Request
-    -> m a
+    -> m (Either String a)
 apiCall parser req = do
     putTextLn $ show req
     v <- getResponseBody <$> httpJSON req
     case parseEither parser v of
-        Right o -> pure o
+        Right o -> pure $ Right o
         Left e -> error $ "couldn't parse response "
```
ksaric: So a `Left` returns what? An error? I think your work here is done, I told you this is an _example_.

src/DataSource/Http.hs
```diff
@@ -303,13 +342,22 @@ iteratePagesWithDelay seconds req = do
...
-            case nextPagen of
-                Just nextUrl -> go (list' <> pagen) nextUrl
-                Nothing      -> pure (list' <> pagen)
-
-    (PageResultList page0 nextPage _) <- liftIO $ apiCall parseJSON req
-    case nextPage of
-        Just nextUrl -> liftIO $ go page0 nextUrl
-        Nothing      -> pure page0
-
+            case req' of
```
ksaric: What's that thingy that can chain computations? Hint: it uses a `bind` operator.

src/DataSource/Http.hs
```diff
@@ -265,14 +295,16 @@ getTicketComments tId = do
 
     let url = showURL $ TicketCommentsURL tId
     let req = apiRequest cfg url
+    case req of
```
ksaric: Chain computation?

src/DataSource/Http.hs
```diff
@@ -240,8 +265,13 @@ _getUser = do
 
     let url = showURL UserInfoURL
     let req = apiRequest cfg url
-
-    apiCall parseJSON req
+    case req of
```
ksaric: Chain computation?

src/DataSource/Http.hs
```diff
@@ -205,8 +223,15 @@ postTicketComment ticketInfo zendeskResponse = do
 
     let responseTicket = createResponseTicket (cfgAgentId cfg) ticketInfo zendeskResponse
     let url  = showURL $ TicketsURL (zrTicketId zendeskResponse)
-    let req = addJsonBody responseTicket (apiRequest cfg url)
-    void $ apiCall (pure . encodeToLazyText) req
+    let req = apiRequest cfg url
+    case req of
```
ksaric: Chain computation?

src/DataSource/Http.hs
```diff
@@ -177,19 +186,28 @@ getExportedTickets time = do
...
-                                        else pure (list' <> pagen)
-
-                    Nothing      -> pure (list' <> pagen)
-
-
-        (PageResultList page0 nextPage _) <- apiCall req
-        case nextPage of
-            Just nextUrl -> go page0 nextUrl
-            Nothing      -> pure page0
+                case req' of
```
ksaric: Chain computation?

src/DataSource/Http.hs
```diff
@@ -82,10 +83,11 @@ getTicketInfo ticketId = do
 
     let url = showURL $ TicketsURL ticketId
     let req = apiRequest cfg url
-
-    apiCall <- asksHTTPNetworkLayer hnlApiCall
-
-    Just <$> apiCall parseTicket req
+    case req of
```
ksaric: Chain computation?

ksaric (changes requested):
Some fixes required.

log-classifier.cabal
mdimjasevic: This is not something I've done by hand. Rather, it was done automatically by a tool while I work on the project, I guess by the editor.

src/HttpLayer.hs
mdimjasevic: There are multiple external functions that get called there and they can fail with an `HttpException` or with a `JSONException`. We shouldn't be hoping it will always be a happy case.

src/HttpLayer.hs
mdimjasevic: I agree with your observation. However, an exception can occur and that's why there is a call to the `catch` function.

src/HttpLayer.hs
mdimjasevic: These two functions have similar, though different signatures:

```haskell
either :: (a -> c) -> (b -> c) -> Either a b -> c
mapLeft :: (a -> c) -> Either a b -> Either c b
```

The `mapLeft` function returns an `Either`, while `either` returns a plain value that erases the information if it is a valid or invalid value.

src/HttpLayer.hs
mdimjasevic: I am sorry, I didn't understand what you said. Can you please rephrase?

src/HttpLayer.hs
ksaric: Where in does `apiRequest` it says that?

src/HttpLayer.hs
ksaric: In most cases, you don't need such functions, if you need to collapse either you use the `either` function, otherwise you can case match. Introducing functions for common operations just brings confusion.

src/HttpLayer.hs
ksaric: And where do you catch it?

ksaric (changes requested):
Some small fixes.

src/DataSource/Http.hs
```diff
@@ -239,9 +259,12 @@ _getUser = do
     apiCall     <- asksHTTPNetworkLayer hnlApiCall
 
     let url = showURL UserInfoURL
-    let req = apiRequest cfg url
-
-    apiCall parseJSON req
+    res <- runEitherT $ do
```
ksaric: You can simplify if you want to use transformers here and remove `transformers-either`:
```
    res <- runExceptT $ do
        req <- ExceptT . pure $ apiRequest cfg url
        ExceptT $ apiCall parseJSON req
```

But I don't think we will require this.

ksaric: `res <- runExceptT $ (ExceptT . pure $ apiRequest cfg url) >>= ExceptT . apiCall parseJSON`

src/HttpLayer.hs
mdimjasevic: There is a `handlerList` in the body of the function providing these two types of exceptions.

src/HttpLayer.hs
mdimjasevic: It's caught right there, on the first line of `addJsonBody`.

src/DataSource/Http.hs
mdimjasevic: Putting it all into a single line doesn't seem like a good idea because it is harder to read.

src/HttpLayer.hs
ksaric: And where in the type signature does it say that this can fail?

ksaric (commented):
Further explanations.

src/DataSource/Http.hs
ksaric: Yes, I know. The reason it's there is because it has nice symmetry, the one that's relevant and tells you how you can get rid of `transformers-either` and where it says that we won't require this is one above.

src/HttpLayer.hs
mdimjasevic: Maybe I'm getting something wrong, but there is this function call: `catches buildRequest handlerList`. In case `buildRequest` results in an exception given in the `handlerList`, it gets handled by a corresponding handler.

src/DataSource/Http.hs
mdimjasevic: Sure, I'll make the change you suggested above.

src/DataSource/Http.hs
ksaric: Means, "use a Monad". You first need to address the other issues.

src/HttpLayer.hs
ksaric: @mdimjasevic Think about what I'm asking here, don't just answer back - what place is that exception coming from? How do you know that this function needs to address this exception? What kind of signature is on this function, and does that tell you something about its behavior?
Think about exceptions - do they occur on pure functions? In which cases? A lot of things to think about, no?

src/HttpLayer.hs
ksaric: Same here, as in "think why you return the `Either`". Why do you return it?

src/HttpLayer.hs
ksaric: Seems like that matches the return type of the function (hint, hint) :smirk:

src/HttpLayer.hs
```diff
@@ -40,41 +42,69 @@ emptyHTTPNetworkLayer = HTTPNetworkLayer
...
-    addRequestHeader "Content-Type" "application/json" $
-    setRequestBasicAuth
-        (encodeUtf8 cfgEmail <> "/token")
-        (encodeUtf8 cfgToken) $
-    parseRequest_ (toString (cfgZendesk <> path))
+apiRequest
+    :: Config
+    -> Text
+    -> Either String Request
+apiRequest Config{..} u = showErr $ catches buildRequest handlerList
```
ksaric: Same thing as https://input-output-rnd.slack.com/archives/CC19D917E/p1534526013000100

ksaric (changes requested):
Fixes.

src/HttpLayer.hs
mdimjasevic: I updated the implementation of this function. Now it is almost the same as the `apiRequest` function, hence it is clear why `Either` in the return type.

src/HttpLayer.hs
mdimjasevic: All of these questions are easy to answer by peeking at the code. Functions from the HTTP library that are called in `addJsonBody` can throw exceptions, hence `addJsonBody` has a corresponding signature. But you must be right, these are great questions.

src/DataSource/Http.hs
mdimjasevic: These computations are chained. If they were not, the code wouldn't compile.

Here we have the resulting monad `m` and the `Either` monad. They could be combined in a monad stack, but given that there is a single point where a failure could happen, a monad stack would be overkill. Don't you agree?

src/HttpLayer.hs
ksaric: > Functions from the HTTP library

Which ones?

mdimjasevic: The `addJsonBody` function's signature follows from your requirement that the `HTTPNetworkLayer.hnlAddJsonBody` field should have a type `hnlAddJsonBody    :: forall a.    (ToJSON a) => a -> Request -> Either String Request`: https://iohk.myjetbrains.com/youtrack/issue/TSD-85#comment=93-26226

ksaric:
Closing in favor of https://github.com/input-output-hk/log-classifier/pull/57

PR #58, opened 2018-08-30
rcohen/tsd-17 - Enable Weeder CI

robcohen:
This PR changes the build process from pure nix/cabal to nix/stack+weeder.
A lot of nix files were removed with this PR. I will need to update the requisite CI-NIX.md, but I can work on that in a separate PR.


disassembler:
I don't know what the requirements are for log-classifier, but one downside of this refactor is builds won't end up in the nix store. Would it be desirable to have the buildkite nix-shell/stack solution just for running weeder and use the existing setup to build the binaries and cache them in the nix store? Granted that's double the time for building.

ksaric:
@disassembler I'll leave that decision in @robcohen hands. 
Seems like there is no easy fix for this issue - ideally we would need to patch https://github.com/ndmitchell/weeder, but that seems like it could take some time.

robcohen:
@disassembler - Unfortunately I think weeder needs to be in control of the build process due to requiring the .hi GHC file output. I believe it may be possible to pass the requisite options to GHC in the cabal file, then have the proper Weeder command line arguments to accomplish what you are talking about. That said, we may as well just add nix support to weeder, as that is the best solution. 

Weeder is built from stack here, and that's unfortunately repeated every build. Would it speed anything up to seperate the buildkite steps into `stack install weeder` and the next step `/var/lib/buildkite-agent/.local/bin/weeder . --build --test && stack test`?

robcohen:
Removed all the weeds but the build-depends: base one, which I believe is a bug in Weeder. Added .weeder.yaml which will ignore this one weed.

disassembler:
I was more referring to should we build the code twice, once using nix-build and once using stack? Again, this is more of if we need cached builds or not for this tool. If we were to apply something similar to cardano (or plutus) I think the answer would be yes, we need to build it twice to ensure our reproducible cached builds are still done.

ksaric (approved):
LGTM.

PR #62, opened 2018-09-13
[TSD-138] Analyze ticket even if already been analyzed

HirotoShioi:
Youtrack Link
https://iohk.myjetbrains.com/youtrack/issue/TSD-138

I've also removed some of the text since it was misleading. (Though classifier processed only X tickets, console state that it will process Y tickets.)


ksaric (approved):
LGTM, but ideally I wouldn't delete that code with version checks. 
I guess we could always get it back via git history.

PR #65, opened 2018-09-18
[TSD-141] Implement JSON parser for upcoming structured logging

HirotoShioi:
Youtrack: https://iohk.myjetbrains.com/youtrack/issue/TSD-141

### DIFF
--- a/src/LogAnalysis/Classifier.hs
+++ b/src/LogAnalysis/Classifier.hs
@@ -69,7 +69,7 @@ extractMessages logfile =
         let decodedLogLines = map eitherDecodeStrict' logLines
 
         when (any isLeft decodedLogLines) $ do
-            let errorText = fromMaybe "Decoding failed" (safeHead $ lefts decodedLogLines)
+            let errorText = fromMaybe "Failed to decode JSON" (safeHead $ lefts decodedLogLines)
             throwM $ JSONDecodeFailure errorText
 
         --  Collect message field since those are the ones we care about
--- a/src/LogAnalysis/Classifier.hs
+++ b/src/LogAnalysis/Classifier.hs
@@ -34,7 +34,7 @@ extractIssuesFromLogs :: (MonadCatch m) => [LogFile] -> Analysis -> m Analysis
 extractIssuesFromLogs logFiles analysis = do
 
     -- Will raise exception if we find an issue.
-    _                   <- mapM checkFile . map getLogFileContent $ logFiles
+    _                   <- mapM (checkFile . getLogFileContent) logFiles
 
     logLines            <- concatMapM extractMessages logFiles
     let analysisResult  = foldl' analyzeLine analysis logLines
--- a/src/LogAnalysis/Classifier.hs
+++ b/src/LogAnalysis/Classifier.hs
@@ -9,17 +9,22 @@ module LogAnalysis.Classifier
        , prettyFormatLogReadError
        , prettyFormatNoIssues
        , prettyFormatNoLogs
+       , runClassifier
+       , runClassifierJSON
        ) where
 
 import           Universum
 
+import           Data.Aeson (eitherDecodeStrict')
+import qualified Data.ByteString.Char8 as C8
 import qualified Data.Map.Strict as Map
 import           Data.Text (isInfixOf)
 import           Data.Text.Encoding.Error (ignore)
 
 import           DataSource.Types (TicketInfo (..), ZendeskAPIUrl (..), showURL)
 import           LogAnalysis.Exceptions (LogAnalysisException (..))
-import           LogAnalysis.Types (Analysis, Knowledge (..), renderErrorCode)
+import           LogAnalysis.Types (Analysis, CardanoLog (..), Knowledge (..), renderErrorCode)
+import           Prelude as P (head)
 
 -- | Number of error texts it should show
 numberOfErrorText :: Int
@@ -28,21 +33,25 @@ numberOfErrorText = 3
 -- | Analyze each log file based on the knowlodgebases' data.
 extractIssuesFromLogs :: (MonadCatch m) => [ByteString] -> Analysis -> m Analysis
 extractIssuesFromLogs files analysis = do
-    analysisResult <- foldlM runClassifiers analysis files
+     -- When 1.4 is out, switch this function with runClasssifierJSON
+    strictLogFiles  <- mapM checkFile files
+    analysisResult <- foldlM runClassifier analysis strictLogFiles
     filterAnalysis analysisResult
-
--- | Run analysis on given file
-runClassifiers :: (MonadCatch m) => Analysis -> ByteString -> m Analysis
-runClassifiers analysis logfile = do
-    -- Force the evaluation of the whole file.
-    strictLogfile <- catchAnyStrict (pure logfile) $ \_ -> throwM LogReadException
-    pure . foldl' analyzeLine analysis . lines . decodeUtf8With ignore $ strictLogfile
   where
-
+    -- Force the evaluation of the whole file.
+    checkFile :: (MonadCatch m) => ByteString -> m ByteString
+    checkFile logfile = catchAnyStrict (pure logfile) $ \_ -> throwM LogReadException
     -- | A helpful utility function.
     -- TODO(ks): Maybe move it to Util?
     catchAnyStrict :: (MonadCatch m, NFData a) => m a -> (SomeException -> m a) -> m a
     catchAnyStrict m = catchAny $ m >>= (return $!) . force
+
+-- | Run analysis on given file
+runClassifier :: (MonadCatch m) => Analysis -> ByteString -> m Analysis
+runClassifier analysis logfile = do
+    let decodedFile = decodeUtf8With ignore logfile
+    let logLines = lines decodedFile
+    pure $ foldl' analyzeLine analysis logLines
 
 -- | Analyze each line
 analyzeLine :: Analysis -> Text -> Analysis
@@ -65,6 +74,20 @@ filterAnalysis as = do
 
 extractErrorCodes :: Analysis -> [Text]
 extractErrorCodes as = map (\(Knowledge{..}, _) -> renderErrorCode kErrorCode) $ Map.toList as
+
+-- | Run analysis on given JSON file
+runClassifierJSON :: (MonadCatch m) => Analysis -> ByteString -> m Analysis
+runClassifierJSON analysis logfile = do
+    let logLines = C8.lines logfile
+    let decodedLogLines = map eitherDecodeStrict' logLines
+
+    when (any isLeft decodedLogLines) $ do
+        let errorText = P.head $ lefts decodedLogLines
+        throwM $ JSONDecodeFailure errorText
+
+    --  Get message field out since those are the ones we care about
+    let messages = map clMessage (rights decodedLogLines)
+    pure $ foldl' analyzeLine  analysis messages
 
 prettyHeader :: Text
 prettyHeader =
--- a/src/LogAnalysis/Classifier.hs
+++ b/src/LogAnalysis/Classifier.hs
@@ -22,21 +22,21 @@ import           Data.Text.Encoding.Error (ignore)
 
 import           DataSource.Types (TicketInfo (..), ZendeskAPIUrl (..), showURL)
 import           LogAnalysis.Exceptions (LogAnalysisException (..))
-import           LogAnalysis.Types (Analysis, CardanoLog (..), Knowledge (..), renderErrorCode)
-import           Prelude as P (head)
+import           LogAnalysis.Types (Analysis, CardanoLog (..), FileFormat (..), Knowledge (..),
+                                    LogFile (..), renderErrorCode)
 
 -- | Number of error texts it should show
 numberOfErrorText :: Int
 numberOfErrorText = 3
 
 -- | Analyze each log file based on the knowlodgebases' data.
-extractIssuesFromLogs :: (MonadCatch m) => [(FilePath, ByteString)] -> Analysis -> m Analysis
+extractIssuesFromLogs :: (MonadCatch m) => [LogFile] -> Analysis -> m Analysis
 extractIssuesFromLogs files analysis = do
 
     -- Check if any of the file is corrupted
-    strictLogFiles <- forM files $ \(path, file) -> do
-        checkedFile <- checkFile file
-        return (path, checkedFile)
+    strictLogFiles <- forM files $ \logfile -> do
+        checkedFile <- checkFile (lfContent logfile)
+        return logfile {lfContent = checkedFile}
 
     logLines       <- concatMapM extractMessages strictLogFiles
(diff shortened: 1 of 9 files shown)


HirotoShioi:
I guess 1 more thing I could do is to wrap the decoded log file with newtype instead of raw `ByteString`

HirotoShioi:
From what I heard from Kristijan, structure of json might have few changes.
Let's not merge this until that is settled.

src/LogAnalysis/Classifier.hs
```diff
@@ -9,41 +9,73 @@ module LogAnalysis.Classifier
...
+import           LogAnalysis.Types (Analysis, CardanoLog (..), Knowledge (..), renderErrorCode)
+import           Prelude as P (head)
 
 -- | Number of error texts it should show
 numberOfErrorText :: Int
 numberOfErrorText = 3
 
 -- | Analyze each log file based on the knowlodgebases' data.
-extractIssuesFromLogs :: (MonadCatch m) => [ByteString] -> Analysis -> m Analysis
+extractIssuesFromLogs :: (MonadCatch m) => [(FilePath, ByteString)] -> Analysis -> m Analysis
```
ksaric: Propagating `FilePath` might be a bit misleading. What we really want to propagate is:
```
data LogFormat = Txt | JSON
```

And also, almost as if `(FilePath, ByteString)` gets propagated through the codebase. As if... As if it's important enough to be wrapped in a newtype :wink:

src/LogAnalysis/Classifier.hs
```diff
@@ -9,41 +9,73 @@ module LogAnalysis.Classifier
...
+-- | Extract messages from either plain or json file
+extractMessages :: (MonadCatch m) => (FilePath, ByteString) -> m [Text]
+extractMessages (path, content) =
+    if toText path `isInfixOf` ".json"
+        then extractMessagesJSON content
+        else extractMessagesPlain content
+  where
+    -- Extract log message from plain log file
+    extractMessagesPlain :: (MonadCatch m) => ByteString -> m [Text]
+    extractMessagesPlain c = do
```
ksaric: What is `c`? Please don't use abbreviations unless you must. And you never really have to.

src/LogAnalysis/Classifier.hs
```diff
@@ -9,41 +9,73 @@ module LogAnalysis.Classifier
...
+        let decodedFile = decodeUtf8With ignore c
+        return $ lines decodedFile
+    -- Extract log messages from JSON log file
+    extractMessagesJSON :: (MonadCatch m) => ByteString -> m [Text]
+    extractMessagesJSON c = do
+        let logLines = C8.lines c
+        let decodedLogLines = map eitherDecodeStrict' logLines
+
+        when (any isLeft decodedLogLines) $ do
+            let errorText = P.head $ lefts decodedLogLines
```
ksaric: `safeHead`, using a total function (if I remember correctly you were saying that non-total functions are pure evil) and then `maybe` or simple case.

src/LogAnalysis/Types.hs
```diff
@@ -94,3 +98,72 @@ instance Eq Knowledge where
 
 instance Ord Knowledge where
     e1 <= e2 = kErrorCode e1 <= kErrorCode e2
+
+-- | Cardano log data type
+data CardanoLog = CardanoLog {
```
ksaric, line 129: Leave `clLog` on but simply put it as optional, using `Maybe`, like `clHost` has to be.

src/Util.hs
```diff
@@ -13,12 +13,12 @@ import           Exceptions (ZipFileExceptions (..))
...
 -- TODO(ks): What happens with the other files? We just ignore them?
-extractLogsFromZip :: Int -> LByteString -> Either ZipFileExceptions [ByteString]
+extractLogsFromZip :: Int -> LByteString -> Either ZipFileExceptions [(FilePath, ByteString)]
 extractLogsFromZip numberOfFiles file = do
     zipMap <- readZip file  -- Read File
-    let extractedLogs :: [ByteString]
-        extractedLogs = map LBS.toStrict . Map.elems . mTake numberOfFiles $ zipMap
-    return extractedLogs
+    let extractedLog = Map.toList $ mTake numberOfFiles zipMap
+    let extractedStrictLog = map (\(path, content) -> (path, LBS.toStrict content)) extractedLog
```
ksaric: Here is a nice place where you can use `where` (`\(path, content) -> (path, LBS.toStrict content)`)and check if the filetype is "txt" or "json".

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+-- | Generate random plain cardano-log file
+genLogFile :: Gen ByteString
+genLogFile = do
+    numOfLines <- choose (1,1000)
+    logLines   <- vectorOf numOfLines (arbitrary :: Gen ByteString)
+    pure $ C8.unlines logLines
+
+-- | Generate error messages that classifier needs to catch
+genErrorText :: Gen Text
+genErrorText = elements
```
ksaric: Nice idea, something I was planning to add as well.

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+    , "irrelevant to given wallet"
+    , "Network.Socket.recvBuf: resource vanished"
+    , "IO error: Failed to create dir"
+    , "open.lock: Locked by"
+    , "Network.Socket.recvBuf: failed (No error)"
+    ]
+
+-- | Generate random knowledge with given text as `kErrorText`
+genKnowledgeWithErrorText :: Text -> Gen Knowledge
+genKnowledgeWithErrorText eText = do
```
ksaric, line 197: I'm sure you thought about that, but it is possible but unlikely, to hit a collision with some existing issue.
You can simply remove those if they ever appear. `... shouldBe . not . elem ... `

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+testExtractIssuesFromLogs knowledge logWithFilePath = do 
+        let analysis = setupAnalysis [knowledge]
+        run $ try $ extractIssuesFromLogs [logWithFilePath] analysis
+
+-- | Formant given UTCTime into ISO8601
+showIso8601 :: UTCTime -> String
+showIso8601 = formatTime defaultTimeLocale "%FT%T%QZ"
+
+-- | Given an list of text, pick and element and encode it
+encodedElements :: [Text] -> Gen ByteString
```
ksaric: Put this in a `where` condition of the function it's used in. Otherwise, people need to think about what it does (on the top level).

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+-- | Generalized testing of extractIssuesFromLogs
+testExtractIssuesFromLogs :: Knowledge 
+                          -> (FilePath, ByteString)
+                          -> PropertyM IO (Either LogAnalysisException Analysis)
+testExtractIssuesFromLogs knowledge logWithFilePath = do 
+        let analysis = setupAnalysis [knowledge]
+        run $ try $ extractIssuesFromLogs [logWithFilePath] analysis
+
+-- | Formant given UTCTime into ISO8601
+showIso8601 :: UTCTime -> String
```
ksaric: Put in `where`.

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+
+-- | Generate random knowledge with given text as `kErrorText`
+genKnowledgeWithErrorText :: Text -> Gen Knowledge
+genKnowledgeWithErrorText eText = do
+    randomKnowledge <- arbitrary
+    pure $ randomKnowledge {kErrorText = eText}
+
+-- | Generate tuple of (FilePath, JSONFile)
+-- with given errorText included as msg
+genJSONWithError :: Text -> Gen (FilePath, ByteString)
```
ksaric: This is where the type description would come in handy. You don't need to deal with `ByteString` but add a `LogType` and you can have a single generator.
Also, remove the `Text` parameter and generate it in the function.

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+        \\"pid\": \""<> show randomPid <> "\",                       \
+        \\"loc\": null,                                              \
+        \\"host\": \""<> randomHost <> "\",                          \
+        \\"sev\": \""<> randomSev <> "\",                            \
+        \\"thread\": \"ThreadId " <> show randomThreadId <> "\"      \
+        \}"
+
+-- | Generate random cardano-log file path
+-- sample: node.json-20180911134009
+genLogJSONFilePath :: Gen FilePath
```
ksaric: Add generator for `LogType` and then you add a single test that tests `FilePath -> LogType` with a single generator.

src/LogAnalysis/Types.hs
```diff
@@ -94,3 +98,72 @@ instance Eq Knowledge where
...
+    , clHost        :: !Text
+    -- ^ Hostname
+    , clSeverity    :: !Text
+    -- ^ Severity of an given log (e.g Info, Notice, Warning, Error)
+    , clThreadId    :: !Text
+    -- ^ Thread id
+    } deriving (Show)
+
+instance FromJSON CardanoLog where
+    parseJSON = withObject "Cardano node log" $ \o -> do
```
ksaric: Are you sure the object is called `"Cardano node log"` ?

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+classifierSpec = do
+    describe "CardanoLog FromJSON" $ modifyMaxSuccess (const 1000) $
+       it "should be able to decode cardano log" $
+           forAll (genLogText Nothing) $ \(logText :: ByteString) -> do
+               let decodedText = eitherDecodeStrict' logText :: Either String CardanoLog
+               isRight decodedText `shouldBe` True
+
+    describe "extractMessages" $ modifyMaxSuccess (const 100) $ do
+        it "should be able to extract log messages from JSON file" $
+            forAll ((,) <$> genLogJSONFilePath <*> genJSONLogFile)
```
ksaric: Please don't use `(,) <$> genLogJSONFilePath <*> genJSONLogFile` - unless it's a very specific distinction that the reader needs to take a look at, I'm not interested in the distinction between `Applicative` and `Monad` and it's killing readability.
For example, I would understand you use this if you have some concurrency code that uses `Applicative` and you want to explain that the things inside it are not related, but in most cases this is not an issue.

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+showIso8601 :: UTCTime -> String
+showIso8601 = formatTime defaultTimeLocale "%FT%T%QZ"
+
+-- | Given an list of text, pick and element and encode it
+encodedElements :: [Text] -> Gen ByteString
+encodedElements xs = encodeUtf8 <$> elements xs
+
+-- | Generate random cardano log
+genLogText :: Maybe Text -> Gen ByteString
+genLogText mErrorText = do
```
ksaric: This is ok, but we are still missing the test for the JSON serialization roundtrip. Shouldn't be too hard to add - you need to test if `decode . encode` is the same thing you sent it to.

src/LogAnalysis/Classifier.hs
```diff
@@ -9,41 +9,73 @@ module LogAnalysis.Classifier
...
+  where
+    -- Extract log message from plain log file
+    extractMessagesPlain :: (MonadCatch m) => ByteString -> m [Text]
+    extractMessagesPlain c = do
+        let decodedFile = decodeUtf8With ignore c
+        return $ lines decodedFile
+    -- Extract log messages from JSON log file
+    extractMessagesJSON :: (MonadCatch m) => ByteString -> m [Text]
+    extractMessagesJSON c = do
+        let logLines = C8.lines c
```
ksaric: This is strange. Something we need to discuss with the guys about the logging format.

test/LogAnalysisSpec.hs
```diff
@@ -0,0 +1,203 @@
...
+    filepath      <- genLogJSONFilePath
+    file          <- genJSONLogFile
+    jsonWithError <- genLogText (Just errorText)
+
+    let fileWithError = jsonWithError <> "\n" <> file
+    pure (filepath, fileWithError)
+
+-- | Generate tuple of (FilePath, LogFile) 
+-- with given errorText included in the ByteString
+genLogWithError :: Text -> Gen (FilePath, ByteString)
```
ksaric: Remove `Text` param and handle both cases. Single generator, test case that covers multiple cases.

ksaric (changes requested):
There are some changes required.

src/LogAnalysis/Classifier.hs
HirotoShioi: I agree. 
In that case, I'll define an datatype `LogFile` which takes `FileFormat` and `ByteString` as argument.

test/LogAnalysisSpec.hs
HirotoShioi: Ah, I see. I can add that.

src/LogAnalysis/Classifier.hs
HirotoShioi: Okay! Please let me know how the discussion goes!

test/LogAnalysisSpec.hs
HirotoShioi: Well, I want to be able to test both cases where I can actually catch the errorText and can't find anything at all which means I need to be able to have some control of how the content of a file should be.
I changed it to take `Maybe Text` as an argument instead. If it's not what you're expecting, then I'll fix it.

test/LogAnalysisSpec.hs
HirotoShioi: Fixed!

test/LogAnalysisSpec.hs
HirotoShioi: Ah I finally understood what you mean. I'll add test next week.

HirotoShioi:
Note to myself.
Need test on `toLogFile` to see if it actually distinguishes the file correctly.

src/LogAnalysis/Types.hs
```diff
@@ -48,6 +56,20 @@ data Knowledge = Knowledge
     -- ^ The FAQ number that will be displayed on the official Cardano FAQ page
     }
 
+-- | File format
+data FileFormat
```
ksaric: `(LogFile lfContent * FileFormat) + (LogFile lfContent * FileFormat)` == `(LogFile lfContent + LogFile lfContent) * FileFormat` == `(LogFile lfContent * (1 + 1)) * FileFormat` == `(2 * LogFile lfContent) * FileFormat` == `2 * (LogFile lfContent * FileFormat)`

ksaric: Will fix/simplify for example.

ksaric (approved):
LGTM.

ksaric:
Merging since I want to rebase my work on top of this to merge all open PRs and move on with our work.

PR #68, opened 2018-09-25
[TSD-144] Improvements on handling corrupted log file.

HirotoShioi:
YT issue: https://iohk.myjetbrains.com/youtrack/issue/TSD-144

It seems like the program fails when we try to evaluate **list** of log files since zip extraction function returns `error` instead of list of files with some of them being an `error`

I did try processing all the previous tickets where we had issues and it was working properly.

### DIFF
--- a/src/Exceptions.hs
+++ b/src/Exceptions.hs
@@ -1,5 +1,6 @@
 module Exceptions
-    ( ProcessTicketExceptions (..)
+    ( ClassifierExceptions (..)
+    , ProcessTicketExceptions (..)
     , ZipFileExceptions (..)
     ) where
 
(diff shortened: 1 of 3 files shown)


src/Exceptions.hs
```diff
@@ -27,8 +28,15 @@ data ZipFileExceptions
     -- ^ Decompresson of a zip file was not sucessful
     deriving Show
 
+-- | Exception on running log-classifier
+data ClassifierExceptions
```
ksaric, line 32: Nice.

src/LogAnalysis/Classifier.hs
```diff
@@ -34,15 +34,15 @@ extractIssuesFromLogs :: (MonadCatch m) => [LogFile] -> Analysis -> m Analysis
 extractIssuesFromLogs logFiles analysis = do
 
     -- Will raise exception if we find an issue.
-    _                   <- mapM (checkFile . getLogFileContent) logFiles
+    _  <- checkFiles $ map getLogFileContent logFiles
```
ksaric, line 37: Is this the culprit then?

ksaric (commented):
Looks good.

src/LogAnalysis/Classifier.hs
HirotoShioi: @ksaric Yes!
Previously, we caught the error by checking each of the file.
But it seems like you cannot catch the error found on TSD-144 by doing that.
I guess the `logFiles` itself is throwing an error so when we try to perform `mapM`, the classifier instantly crashes because `logFiles` itself is an error.

ksaric (approved):
LGTM.

PR #72, opened 2018-10-02
[TSD-143] Create the web application to enable other users to use it more easily

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-143

Also removing Weeder since it was causing some issues when building. Tried to debug with @robcohen.

Ready for review.

If you want to see how it works, you first need to create `app_users.json` in the directory you are starting the server from and put some users inside. 
For example - `[ { "username":"ksaric","password":"pass" } ]`


ksaric:
This stayed without any review for too long, I need to continue the work using this and so I merged it.

PR #73, opened 2018-10-04
Added a number of new entries in CSV and modified types

robcohen:
Added to CSV all IOHKS tickets with YT filter: Zendesk-Debugger: Yes State: -Done
Added datatypes for new additions in CSV.


src/LogAnalysis/KnowledgeCSVParser.hs
```diff
@@ -30,20 +30,27 @@ quotedText = do
...
-    <|> (string "DBPath"            >> return DBPath)
-    <|> (string "CannotGetDBSize"   >> return CannotGetDBSize)
-    <|> (string "DBError"           >> return DBError)
-    <|> (string "BalanceError"      >> return BalanceError)
-    <|> (string "NetworkError"      >> return NetworkError)
-    <|> (string "ConnectionRefused" >> return ConnectionRefused)
-    <|> (string "ResourceVanished"  >> return ResourceVanished)
-    <|> (string "Unknown"           >> return Unknown)
-    <|> (string "Error"             >> return Error)
+        (string "TLSCert"   >> return TLSCert)
```
ksaric: Align this, please, hard to read.

src/Lib.hs
```diff
@@ -225,6 +225,7 @@ fetchAgents dataLayer = do
 
 -- | 'processTicket' with exception handling
 processTicketSafe :: DataLayer App ->TicketId -> App ()
+--processTicketSafe dataLayer tId = catch (putTextLn . show =<< (processTicket dataLayer tId))  -- Debug output
```
ksaric: You really don't need this in the code anymore.

src/Lib.hs
```diff
@@ -254,7 +255,7 @@ processTicket dataLayer tId = do
             zendeskResponse <- getZendeskResponses dataLayer comments attachments ticketInfo
 
             -- post ticket comment
-            postTicketComment ticketInfo zendeskResponse
+            --postTicketComment ticketInfo zendeskResponse
```
ksaric: Restore posting, default behavior.

src/LogAnalysis/Types.hs
```diff
@@ -100,20 +107,28 @@ instance Show Knowledge where
 -- | Sorted accoring to knowledgebase.
 -- Tag needs to be in lowercase since Zendesk automatically convert any uppercase
 -- lowercase
+
 renderErrorCode :: ErrorCode -> Text
-renderErrorCode DBError              = "db-corrupted"
-renderErrorCode StaleLockFile        = "stale-lock-file"
+renderErrorCode TLSCert              = "tls-cert-error"
```
ksaric, line 116: I still don't understand why you removed some of the existing entries. You said that there weren't any of them in the ticket tags, but that doesn't mean that there won't be any of them.

And it begs the questions where is the knowledge base coming from - from our CSV/YouTrack or your head? If it's coming from your head then that is no good. If you don't have them in YouTrack, please write down (in YouTrack) which ones have been removed and why, since only you know that now.

ksaric (changes requested):
Small fixes and a question to be answered.

src/LogAnalysis/Types.hs
robcohen: The CSV is generated from YouTrack. i export the CSV, then transform it using the python script I wrote.

robcohen: Re-added the entries I previously removed.

ksaric (approved):
Ok, its in your hands.

PR #74, opened 2018-10-05
[TSD-150] Build failing when using just Nix.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-150

Fixed building from scratch with Nix.


ksaric:
We probably need to merge https://github.com/input-output-hk/log-classifier/pull/72 first since that fixes the `weeder` issue.

shell.nix
```diff
@@ -3,11 +3,11 @@ let
...
         overrides = haskellPackagesNew: haskellPackagesOld: rec {
-          universum =
-            pkgs.haskell.lib.dontCheck
-              (haskellPackagesNew.callHackage "universum" "1.1.0" {});
-          log-classifier =
-            (haskellPackagesNew.callCabal2nix "log-classifier" ./. {});
+          universum           = pkgs.haskell.lib.dontCheck (haskellPackagesNew.callHackage "universum" "1.1.0" {});
+          concurrent-output   = pkgs.haskell.lib.dontCheck (haskellPackagesNew.callPackage ./concurrent-output.nix {});
+          # because of http://hackage.haskell.org/package/concurrent-output
+          stm                 = pkgs.haskell.lib.dontCheck (haskellPackagesNew.callPackage ./stm.nix {});
```
robcohen, line 9: Would rather just have CallHackage on everything to reduce unnecessary nix files. Tried to fix it and got an infinite recursion dependency issue. Will fix later.

robcohen (approved):
LGTM

shell.nix
ksaric: That's what I got as well so I decided to use this instead. AFAICU, infinite recursion dependency issue occurs when there is a namespace collision?

PR #76, opened 2018-10-09
[TSD-148] Refactor inspectAttachment

HirotoShioi:
WIP!

YouTrack: https://iohk.myjetbrains.com/youtrack/issue/TSD-148

### DIFF
--- a/classifier/src/Lib.hs
+++ b/classifier/src/Lib.hs
@@ -30,11 +30,8 @@ import           Universum
 
 import           UnliftIO.Async (mapConcurrently)
 
-<<<<<<< 88c4cea847162e0760fdd3fcff7ffd2521561c2f:classifier/src/Lib.hs
-=======
 import           Control.Exception.Safe (Handler (..), catches)
 import           Data.Attoparsec.Text.Lazy (eitherResult, parse)
->>>>>>> [TSD-148] Refactor inspectAttachment (WIP):src/Lib.hs
 import qualified Data.ByteString.Lazy as BS
 import           Data.Attoparsec.Text.Lazy (eitherResult, parse)
 import           Data.List (nub)
@@ -237,7 +234,8 @@ processTicket dataLayer tId = do
 
             -- post ticket comment
             -- Maybe for now we don't need to actually post this, but let the agent post it.
-            --postTicketComment ticketInfo zendeskResponse
+           -- postTicketComment ticketInfo zendeskResponse
+            print zendeskResponse
 
             pure zendeskResponse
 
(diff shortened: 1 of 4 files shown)


classifier/src/Lib.hs
```diff
@@ -363,68 +369,67 @@ inspectAttachments dataLayer ticketInfo attachments = do
...
-            tryAnalysisResult    <- try $ extractIssuesFromLogs logFiles analysisEnv
+inspectAttachment :: (MonadCatch m)
+                  => Config
+                  -> TicketInfo
+                  -> AttachmentContent
+                  -> ExtractLogFileFunc
+                  -> ExtractErrorCodeFunc m
+                  -> m ZendeskResponse
+inspectAttachment Config{..} ticketInfo attachment extractLogFileFunc extractIssueFunc =
+    flip catches [Handler handleZipFileException, Handler handleLogAnalysisException] $ do
```
ksaric, line 392: https://github.com/fpco/safe-exceptions/blob/master/src/Control/Exception/Safe.hs#L265

classifier/src/Lib.hs
```diff
@@ -363,68 +369,67 @@ inspectAttachments dataLayer ticketInfo attachments = do
...
-    inspectAttachment config ticketInfo att
+    inspectAttachment config ticketInfo att extractLogsFromZip extractIssuesFromLogs
   where
     handleMaybe :: Maybe a -> App a
     handleMaybe Nothing  = throwM $ AttachmentNotFound (tiId ticketInfo)
     handleMaybe (Just a) = return a
 
+-- This is to enable test cases on 'inspectAttachment'
+-- Perhaps we can add this to 'Config' ?
+type ExtractLogFileFunc = Int -> LByteString -> Either ZipFileExceptions [LogFile]
```
ksaric, line 379: This is nice as a first approximation. In the long run, it would be better to put these functions into a "layer" as we did so far since we (usually) pass these functions together in a record because they are easy to think about and organize code like that.

classifier/src/Lib.hs
```diff
@@ -363,68 +369,67 @@ inspectAttachments dataLayer ticketInfo attachments = do
...
-                                { zrTicketId    = tiId
-                                , zrComment     = prettyFormatNoIssues ticketInfo
-                                , zrTags        = TicketTags [renderTicketStatus NoKnownIssue]
-                                , zrIsPublic    = cfgIsCommentPublic
-                                }
-                        JSONDecodeFailure errorText ->
-                            -- Need to respond to an ticket instead of throwing exception
-                            throwM $ JSONDecodeFailure errorText
+     mkZendeskErrorResponse :: Text -> ErrorCode -> ZendeskResponse
+     mkZendeskErrorResponse comment e = mkZendeskResponse comment [renderErrorCode e]
```
ksaric: `e`? Don't use abbreviations.

ksaric (dismissed):
LGTM, a couple of minor comments.

classifier/src/Lib.hs
HirotoShioi: Fixed!

classifier/src/Lib.hs
HirotoShioi: I agree. Should I make a new layer and add it to `Config` ?

classifier/src/Lib.hs
HirotoShioi: Well, since I'm catching multiple exceptions, I'd need `handles` function which should be

```
handles :: (C.MonadCatch m, C.MonadThrow m) => [Handler m a] -> m a-> m a
handles handlers io = flip catches handlers io
```

ksaric (approved):
LGTM.

PR #77, opened 2018-10-09
[TSD-149] Refactor project into submodules.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-149

Fixing the project structure so we can start adding the frontend.


web/src/Main.hs
```diff
@@ -0,0 +1,15 @@
+module Main where
+
+import           Universum
+
+import           DataSource (TicketId (..), runApp)
+import           Lib (createBasicDataLayerIO, createConfig, exportZendeskDataToLocalDB, fetchAgents,
+                      fetchAndShowTickets, fetchAndShowTicketsFrom, fetchTickets,
+                      inspectLocalZipAttachment, processTicketSafe, processTickets,
+                      processTicketsFromTime, showStatistics)
```
HirotoShioi: I don't think these imports are needed

HirotoShioi (commented):
When I run test on local machine with `stack test --fast`, the results are not showing on the console, but instead written in log file. Anyway I can see the result on the console?

web/src/Main.hs
ksaric: Correct. Me and imports :grin:

ksaric:
@HirotoShioi See - https://github.com/commercialhaskell/stack/blob/master/doc/build_command.md#build-output
You can use something like `stack test log-classifier:log-classifier-test` to build and test a single module that will show output.
Also, for now, `stack test --dump-logs` is an option.

HirotoShioi (approved):
Thank you for the changes. I'll approve it.

PR #78, opened 2018-10-16
Add endpoints for fetching/processing tickets

HirotoShioi:
https://iohk.myjetbrains.com/youtrack/issue/TSD-154

Adding endpoints to the web API.

### DIFF
--- a/web/src/Server.hs
+++ b/web/src/Server.hs
@@ -98,8 +98,8 @@ type API
     =  BasicAuthURL :> "api" :> "v1" :> "tickets" :> Capture "ticketId" TicketId :> Get '[JSON] TicketInfo
   :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> Get '[JSON] [TicketInfo]
   :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> Capture "ticketId" TicketId :> "comments" :> Get '[JSON] [Comment]
+  :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> "analysis" :> Post '[JSON] [ZendeskResponse]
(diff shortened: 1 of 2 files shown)


web/src/Server.hs
```diff
@@ -96,8 +96,10 @@ type BasicAuthURL = BasicAuth "log-classifier" User
 
 type API
     =  BasicAuthURL :> "api" :> "v1" :> "tickets" :> Capture "ticketId" TicketId :> Get '[JSON] TicketInfo
+  :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> Get '[JSON] [TicketInfo]
   :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> Capture "ticketId" TicketId :> "comments" :> Get '[JSON] [Comment]
   :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> "analysis" :> ReqBody '[JSON] CTicketId :> Post '[JSON] ZendeskResponse
+  :<|> BasicAuthURL :> "api" :> "v1" :> "tickets" :> "analysis" :> Get '[JSON] [ZendeskResponse]
```
HirotoShioi: The current issue I have is that I can't use `Post` on processing multiple tickets since when I try to access them via `localhost:4000/api/v1/tickets/analysis`, I get an error message of `not enough inputs`
Is there any workaround for this? I'm currently looking through servant docs regarding this.

ksaric: Ah. I see what you want to do with this. Try moving your endpoint up and move the endpoint with a single `CTicketId` below that.

ksaric: Also, it's not a `Get`, right? We need a `Post` there.

HirotoShioi: Yes, it's supposed to be `Post`! I'll fix it right away!

ksaric (approved):
LGTM.

PR #81, opened 2018-10-18
[TSD-147] - Bash commands to create CSV

robcohen:
This PR consists of the bash commands required to pull the YouTrack JSON and transform it with jq into a CSV.

Note: This outputs the following format:
```
"$CLASSIFIER_TEXT","ISSUE_ID"
```

Please advise whether FAQ, Problem, solution should be added.


scripts/youtrack2csv.py
```diff
@@ -0,0 +1,9 @@
+import pandas as pd
```
ksaric, line 1: This might confuse people at some point, offering two methods while JQ can do the job might be redundant.

ksaric (approved):
LGTM, but I would still remove the Python script.

scripts/youtrack2csv.py
robcohen: That's true, but the python script is much more user friendly to reorganize. JQ is substantively harder to use. Either way, the script will be useless pretty quickly as IOHKS tickets are no longer containing the relevant log-classifier text data and will soon not exist.

PR #83, opened 2018-10-29
[TSD-157] Add environment variable.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/TSD-157

Add environment variables for running with arbitrary files location.
Tested with `LCPATH=/home/ksaric/projects/haskell/log-classifier/tmp-secrets stack exec log-classifier-web --no-nix-pure`.
_Do note the `--no-nix-pure` part!_

One would need these files there:
- token
- assign_to
- knowledge.csv
- app_users.json


robcohen (approved):
Works for me. No recommendations for edits.

